PR #36, opened 2015-11-23
Add scala as supported language

Tarrasch:
For now just run Scala in interpreted mode. Future work could be to use
scalac. I didn't do it in this patch as I'm not familiar enough with the
typical scalac flags.


austrin:
@Tarrasch I'm not sure how useful this is on its own, since there is no Scala support in Kattis.  This would let you use Scala for validators though, is that what you want?

Tarrasch:
I didn't consider that.

But yes. My use case is that I wrote my own reference solution in Scala. And I would like for problemstools to be able to verify my solution against the provided `.ans` files. :)

Tarrasch:
Can this be either merged, closed or reviewed again please?

austrin:
@Tarrasch sorry for lack of response, I thought I had already responded.  I'll close this for now, it is not that generally useful since Kattis does not support Scala.

Tarrasch:
Better late than never.

Though I don't understand why this is not merged as it makes perfect sense to have in my work-flow. But if you think my use-case is rare, I trust you to have made the right decision to close this. :)

PR #44, opened 2016-06-02
Return error when input validator does not compile.

meisterT:
This is relevant if problemparts include 'data' but not 'validators'.

### DIFF
--- a/src/verifyproblem.py
+++ b/src/verifyproblem.py
@@ -674,6 +674,7 @@ class InputFormatValidators(ProblemAspect):
             try:
                 if not val.compile():
                     self.error('Compile error for input format validator %s' % val.name)
+                    self._validators.remove(val)
             except ProgramError as e:
                 self.error(e)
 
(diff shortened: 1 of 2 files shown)


src/verifyproblem.py
```diff
@@ -696,7 +696,9 @@ def validate(self, testcase):
                     testcase.error('Input format validator %s crashed on input %s' % (val, testcase.infile))
                 if os.WEXITSTATUS(status) != 42:
                     testcase.error('Input format validator %s did not accept input %s, exit code: %d' % (val, testcase.infile, os.WEXITSTATUS(status)))
-
+            else:
+                self.error('Compile error for input format validators %s' % val.name)
+                self._validators.remove(val)
```
austrin: Nitpick: `validators` should be `validator`.

A bit more substantial: I guess now one will get the "compile error" twice if running a complete check?  I would be inclined to instead just call `self.check()` before entering the for loop, to rely on the existing code.

meisterT: Good point, I've updated the pull request to run self.check() instead.

PR #46, opened 2016-09-08
Changing from dmcs to mcs to compile C#

pehrsoderman:
`dmcs` is a wapper script, which forces the use of mono 4.0. Using `dmcs` creates a semi-broken installation, as mono 4.5 is shipped with ubuntu 16.04. Therefore we change to use `mcs` instead.

This matches the https://github.com/Kattis/kattis/pull/1974 pull request for kattis backend.

### DIFF
--- a/problemtools/languages.yaml
+++ b/problemtools/languages.yaml
@@ -110,7 +110,7 @@ csharp:
     name: 'C#'
     priority: 700
     files: '*.cs'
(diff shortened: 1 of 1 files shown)


austrin:
@pehrsoderman looks good, but please open it into the `develop` branch rather than `master`.  (Am trying new workflow with releases on `master` and work in progress on `develop`.)

austrin:
@pehrsoderman I've updated develop to be the default branch now so should work better in the future. Sorry for the added trouble.

Issue #34, opened 2015-11-07
Better support of .timelimit file


austrin:
If a problem has a .timelimit file, then verifyproblem should use the time limit stated there if it is within reasonable margin of the calculated (non-rounded) time limit (e.g. if it is within 10% or something like that?).

If it deviates by more, a warning should be issued, I guess.

.timelimit should maybe also be written by verifyproblem?

austrin:
As part of doing this, remember to check that the .timelimit is an integer since other parts of problemtools won't work with a non-integer limit.

niemela:
I would argue that this is not needed after https://github.com/Kattis/problem-package-format/pull/49

PR #54, opened 2016-10-21
Support floats in .timelimit

simonlindholm:
Not sure if it's supported well by other parts of the toolchain, but we had a .timelimit (for the sake of CMS) of 1.5 in [one problem of KATT last year](https://github.com/oskmeister/po2016/blob/master/katt/Contest1/cities/), and problem2pdf choked on it.

### DIFF
--- a/problemtools/templates/latex/problemset.cls
+++ b/problemtools/templates/latex/problemset.cls
@@ -186,7 +186,7 @@
 }
 
 \newcommand{\ps@formattime}[1]{
-  #1\ifnum#1=1 second \else seconds \fi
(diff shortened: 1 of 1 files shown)


simonlindholm:
Flagging for review just because it's a one-line patch that's been around for half a year. (Even if non-integers are not supported by kattis, that's better stated by other tools than pdflatex's contextless "! Missing number, treated as zero.")

austrin:
OK, I guess since the .timelimit part is pretty divorced from verifyproblem at the moment, it is fine.  Added a comment in #34 about dealing with this when integrating .timelimit into verifyproblem.

PR #56, opened 2016-11-26
Fix python2 calls and shebangs

t-wissmann:
Explicitely call `python2` instead of 'python' since on some systems (e.g.
arch linux) python3 is the default version of python. Also do this in
shebangs.

Furthermore in shebangs, use the wrapper `/usr/bin/env` instead
of hard-coding the interpreter path. This makes scripts more portable,
because on some systems (mostly BSDs), bash is not in `/bin/` but
somewhere else in PATH.

See also some stackoverflow references for [Python shebangs](http://stackoverflow.com/a/19305076/4400896) and [Bash shebangs](http://stackoverflow.com/a/10383546/4400896).

### DIFF
--- a/admin/make_release.sh
+++ b/admin/make_release.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e
 
--- a/admin/update_version.py.sh
+++ b/admin/update_version.py.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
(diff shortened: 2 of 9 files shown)


austrin:
@t-wissmann looks good, thanks!  One small comment: there are still two `exec python` left in `bin/problem2html.sh` and `bin/verifyproblem.sh`, would you mind updating those to `python2` as well?

t-wissmann:
Thank you very much! I've updated those two files as well and updated the commit.

PR #60, opened 2017-04-03
Add support for attachments

jsannemo:
The spec[1] states that attachments to the statement can be added to the attachments/ folder.
This commits add support for this in problemtools, and exposes it as part of the statement object.
Directories will be downloaded as zip files, with a .zip extension.
Thus, verification is needed to ensure that the corresponding file does not also exist as attachment.

[1] http://problemarchive.org/wiki/index.php/Problem_Format

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -644,6 +644,40 @@ class ProblemStatement(ProblemAspect):
         return ret
 
 
+class Attachments(ProblemAspect):
+    """Represents the attachments of a problem.
+
+    Attributes:
+        attachments: The absolute paths to the attachment files for this problem.
+
+    """
+
+    def __init__(self, problem):
+        attachments_path = os.path.join(problem.probdir, 'attachments')
+        if os.path.isdir(attachments_path):
+            self.attachments = [os.path.join(attachments_path, attachment_name) for attachment_name in os.listdir(attachments_path)]
+        else:
+            self.attachments = []
+        self.debug('Adding attachments %s' % str(self.attachments))
+
+    def check(self, args):
+        if self._check_res is not None:
+            return self._check_res
+        self._check_res = True
+
+        for attachment_path in self.attachments:
+            if os.path.isdir(attachment_path):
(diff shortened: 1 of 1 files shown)


jsannemo:
Added new class for attachments.

I still added this to the problem statement aspect. My experience is that attachments are very tightly associated with statements (an attachment has always been referred to from the statement when I have seen them used.), so it makes sense to always update them in tandem.

niemela:
> My experience is that attachments are very tightly associated with statements (an attachment has always been referred to from the statement when I have seen them used.), so it makes sense to always check them in tandem (and later, in addproblem, add them).

I agree.

problemtools/verifyproblem.py
```diff
@@ -582,6 +582,13 @@ def __init__(self, problem):
             self.languages.append('')
         for f in glob.glob(glob_path + '[a-z][a-z].tex'):
             self.languages.append(re.search("problem.([a-z][a-z]).tex$", f).group(1))
+        attachments_path = os.path.join(problem.probdir, 'attachments')
+        if os.path.isdir(attachments_path):
+            self.attachments = os.listdir(attachments_path)
+        else:
+            self.attachments = []
+        self.debug('Adding attachments %s' % str(self.attachments))
+
```
austrin: This (and the changes to the check) method should now be removed since the code has moved to the new class.

problemtools/verifyproblem.py
```diff
@@ -644,6 +656,35 @@ def get_config(self):
...
 
 
+class Attachments(ProblemAspect):
+    def __init__(self, problem):
+        self._problem = problem
+        attachments_path = os.path.join(problem.probdir, 'attachments')
+        if os.path.isdir(attachments_path):
+            self.attachments = os.listdir(attachments_path)
+        else:
+            self.attachments = []
```
austrin: It seems like every time the attachments field is used, it is joined with the path of the problem to get an absolute path.  With that in mind it would probably make sense to store full paths in the attachments field instead?

PR #61, opened 2017-04-03
Add support for Pascal

jsannemo:
With the impending Kattis support for pascal, problem tools should also support Pascal.


problemtools/languages.yaml
```diff
@@ -194,3 +194,10 @@ scala:
     files: '*.sc *.scala'
     compile: '/usr/bin/scalac -encoding UTF-8 -sourcepath {path} -d {path} {files}'
     run: '/usr/bin/scala -J-Xmx{memlim}m -classpath {path} {mainclass}'
+
+pascal:
```
niemela: Languages should be listed alphabetically (on the key).

jsannemo: fixed

problemtools/languages.yaml
```diff
@@ -149,6 +149,13 @@ objectivec:
     compile: '/usr/bin/gcc -O2 -std=gnu99 {files} -lm -lobjc -MMD -MP -DGNUSTEP -DGNUSTEP_BASE_LIBRARY=2 -DGNU_GUI_LIBRARY=1 -DGNU_RUNTIME=1 -fno-strict-aliasing -fexceptions -fobjc-exceptions -D_NATIVE_OBJC_EXCEPTIONS -fPIC -Wall -DGSWARN -DGSDIAGNOSE -Wno-import -g -O2 -fstack-protector --param=ssp-buffer-size=4 -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -Werror=format-security -fgnu-runtime -fconstant-string-class=NSConstantString -I/usr/local/include/GNUstep -I/usr/include/GNUstep -lgnustep-base -o {binary}'
     run: '{binary}'
 
+pascal:
+    name: 'Pascal'
+    priority: 350
+    files: '*.pas'
+    compile: '/usr/bin/fpc -o"{mainfile}.out" -O2 -XS -Xt "{mainfile}"'
+    run: '"{mainfile}.out"'
```
austrin: Why not use `{binary}` to be consistent with all the other compiled languages?

jsannemo: I could (which would also require relaxing the "only one entry point meta variable" constraint). I don't really have an opinion.

PR #63, opened 2017-04-18
Add support for new grading settings.

jsannemo:
Some grading settings were moved from problem.yaml to testdata.yaml.
These settings should be read from there instead of problem.yaml.
Grading has also been reworked to accomodate for the fact that every
solution should now use the default grader (regardless of scoring), and
an on_reject setting per test case isntead.

@austrin I've done some prelim tests on old NCPC problems and our old PO problems, which both behave correctly. I also tested a new scoring problem using per-testgroup settings and on_reject, which worked fine.

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -43,43 +43,6 @@ class SubmissionResult:
         self.ac_runtime = -1.0
         self.ac_runtime_testcase = None
 
-
-    @staticmethod
-    def aggregate_results(sub_results, policy, grade=None):
-        res = SubmissionResult(None)
-
-        for r in sub_results:
-            if r.runtime > res.runtime:
-                res.runtime = r.runtime
-                res.runtime_testcase = r.runtime_testcase
-            if r.ac_runtime > res.ac_runtime:
-                res.ac_runtime = r.ac_runtime
-                res.ac_runtime_testcase = r.ac_runtime_testcase
-
-        verdict_value = {'JE': -1, 'CE': 0, 'TLE': 1, 'RTE': 2, 'WA': 3, 'AC': 4}
-
-        rejection = next((r for r in sub_results if r.verdict == 'JE'), None)
-        if rejection is None:
-            if policy == 'first_error':
-                rejection = next((r for r in sub_results if r.verdict != 'AC'), None)
-            elif policy == 'worst_error':
-                rejection = min(sub_results, key=lambda r: verdict_value[r.verdict])
-            # else policy is 'grade' and we should grade the results
-
-        if rejection is None:
-            if grade is not None:
-                (res.verdict, res.score) = grade(sub_results)
-            else:
-                res.verdict = 'AC'
-        else:
-            res.verdict = rejection.verdict
-            res.reason = rejection.reason
-            res.testcase = rejection.testcase
-
-        return res
-
-
-
     def __str__(self):
         verdict = self.verdict
         details = []
@@ -196,9 +159,9 @@ class TestCase(ProblemAspect):
                                       timelim=timelim_high+1,
                                       memlim=self._problem.config.get('limits')['memory'])
             if is_TLE(status) or runtime > timelim_high:
-                res2 = SubmissionResult('TLE', score=self._problem.config.get('grading')['reject_score'])
+                res2 = SubmissionResult('TLE', score=self.testcasegroup.config['reject_score'])
             elif is_RTE(status):
-                res2 = SubmissionResult('RTE', score=self._problem.config.get('grading')['reject_score'])
+                res2 = SubmissionResult('RTE', score=self.testcasegroup.config['reject_score'])
             else:
                 res2 = self._problem.output_validators.validate(self, outfile)
             res2.runtime = runtime
@@ -207,7 +170,7 @@ class TestCase(ProblemAspect):
         if res2.runtime <= timelim_low:
             res1 = res2
         else:
-            res1 = SubmissionResult('TLE', score=self._problem.config.get('grading')['reject_score'])
+            res1 = SubmissionResult('TLE', score=self.testcasegroup.config['reject_score'])
         res1.testcase = res2.testcase = self
         res1.runtime_testcase = res2.runtime_testcase = self
         res1.runtime = res2.runtime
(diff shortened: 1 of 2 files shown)


problemtools/verifyproblem.py
```diff
@@ -293,6 +284,15 @@ def get_subgroup(self, name):
         return next((sub for sub in self._items if isinstance(sub, TestCaseGroup) and os.path.basename(sub._datadir) == name), None)
 
 
+    def get_grader_flags_for_judging(self):
+        problem_type = self._problem.config.get('type')
+        if self.config['grading'] == 'default':
+            # The default grader flags are different depending on the problem type, which requires this override
+            default_grader_spec = ['always_accept', 'sum'] if problem_type == 'scoring' else ['no_errors', 'sum']
+            return default_grader_spec + self.config['grader_flags'].split()
```
austrin: I wonder if this was a bad choice in the spec.  What do you think about letting the default for default_grader always be `no_errors`?  It means one has to set it in the root testdata.yaml for all IOI problems, but on the other hand one gets rid of this annoying special casing, and can also put the default value in the default_grader instead of requiring the CCS to pass a verdict aggregation type.

jsannemo: Yes, I agree very much. @niemela ?

problemtools/verifyproblem.py
```diff
@@ -528,11 +555,16 @@ def check(self, args):
         elif self._data['license'] == 'unknown':
             self.warning("License is 'unknown'")
 
-        if not self._data['grading']['on_reject'] in ['first_error', 'worst_error', 'grade']:
-            self.error("Invalid value '%s' for on_reject policy" % self._data['grading']['on_reject'])
+        if 'on_reject' in self._data['grading']:
+            self.warning('on_reject in problem.yaml is deprecated, use on_reject in testdata.yaml instead.')
```
austrin: Why not add this one to the list of other deprecation checks below?  Doesn't make a big difference but it keeps deprecation checks in one place plus removes the inconsistency in the wording of the message.

jsannemo: Fixed

problemtools/verifyproblem.py
```diff
@@ -310,6 +310,23 @@ def check(self, args):
...
+            # Check grading
+            try:
+                score_range = self.config['range']
+                (min_score, max_score) = map(float, score_range.split())
+                if min_score >= max_score:
+                    self.error("Invalid score range '%s': minimum score must be smaller than maximum score" % score_range)
+            except VerifyError:
+                raise
+            except:
+                self.error("Invalid format '%s' for grading.range: must be exactly two floats" % score_range)
```
austrin: I think there are some checks missing here.  At the very least, should check that on_reject is break or continue.  Also, what happens if I don't specify all the values in `_SCORING_ONLY_KEYS` (for a scoring problem)?

jsannemo: True, fixed.

Those keys have a default value as well, which will be used.

problemtools/verifyproblem.py
```diff
@@ -245,11 +215,32 @@ def __init__(self, problem, datadir, parent=None):
...
+        problem_grading = problem.config.get('grading')
+        for key in ['accept_score', 'reject_score', 'range']:
+            if key in problem.config.get('grading'):
+                self.config[key] = problem_grading[key]
+
+        problem_on_reject = problem_grading.get('on_reject')
+        if problem_on_reject == 'first_error':
+            self.config['on_reject'] = 'break'
+        if problem_on_reject == 'grade':
+            self.config['on_reject'] = 'continue'
```
austrin: For these deprecated properties, maybe only take the data from problem.yaml for the root group?  (It should be the same but matches more closely how at least I think about it.)

Also, maybe I'm reading too sloppily but it looks like any value specified in problem.yaml will override the values specified in testdata.yaml?  The problem.yaml values should only be used if there was nothing specified in testdata.yaml.

jsannemo: OK. Although it should probably be an error rather to have both specified?

problemtools/verifyproblem.py
```diff
@@ -245,11 +215,32 @@ def __init__(self, problem, datadir, parent=None):
...
+
+        if self._problem.config.get('type') == 'pass-fail':
+            for key in TestCaseGroup._SCORING_ONLY_KEYS:
+                if key not in self.config:
+                    self.config[key] = None
+
```
austrin: Possibly simpler to change RHS in the assignment in the next for loop to `default if probtype != 'pass-fail' or field not in TestCaseGroup._SCORING_ONLY_KEYS else None`?  Not sure, am maybe just looking for an excuse to also ask you to fix the misplaced `not` in the next for loop...

problemtools/verifyproblem.py
austrin: Yes, raising an error when both are specified is probably reasonable.  It may be a bit wonky to do that though, since you can't raise errors from the constructor.

But actually, I think I take it back, it's probably better to use the values from problem.yaml everywhere if they are specified, since that's the current behaviour.

In other words, ignore all of this comment.

problemtools/verifyproblem.py
jsannemo: it seems a bit ugly to use to different ways of assigning defaults, imo. also, that becomes a very long expression...

austrin: But but but, fewer loops... Fair enough, point taken.

problemtools/verifyproblem.py
```diff
@@ -361,29 +381,45 @@ def check(self, args):
...
+        judge_error = next((r for r in sub_results if r.verdict == 'JE'), None)
+        if judge_error:
+            res.verdict = judge_error.verdict
+            res.reason = judge_error.reason
+            res.testcase = judge_error.testcase
+        else:
+            (res.verdict, score) = self._problem.graders.grade(sub_results, self, shadow_result)
+            if self._problem.config.get('type') == 'scoring':
+                res.score = score
+            res.testcase = sub_results[-1].testcase
```
austrin: I guess this is a new heuristic for figuring out which test case to "blame" for a result.  For ACM style problems it's going to be a correct heuristic, whereas for other types of problem the quality of the guess may be so-so, but on the other hand we can't really give a more qualified guess.  Have I understood correctly?

problemtools/verifyproblem.py
jsannemo: OK. Ignoring.

problemtools/verifyproblem.py
jsannemo: Precisely so.

problemtools/verifyproblem.py
austrin: I think he will agree, so please go ahead and do this change.

problemtools/verifyproblem.py
jsannemo: Essentially it ensures res.testcase is always the last testcase ran for a certain subresult, which imho seems like a reasonable definition for what the a testcase property for a result would be.

problemtools/verifyproblem.py
austrin: Looks good and I've been testing stuff, feel free to merge after this has been done.

austrin:
Just to make sure it didn't drown: feel free to merge after https://github.com/Kattis/problemtools/pull/63#pullrequestreview-33555767 is done.

jsannemo:
@austrin can you make a release afterwards and bump the kattis version (this will be needed for the addproblem changes)?

PR #67, opened 2017-08-24
Disable pybuild proxy for github.com

jsannemo:
By default, pybuild proxies all http[s] requests to 127.0.0.1:9 to
prevent downloading dependencies from PyPi
(https://wiki.debian.org/Python/Pybuild). This has the consequence of
making the clone of checktestdata fail unless it has been cloned
before (e.g. by running make support previously). Adding a rule to
exclude github.com from this proxy makes building work again.


niemela:
@austrin Can (should) we merge this?

PR #68, opened 2017-09-03
Stub support for kotlin (jvm) target

hex539:
Kotlin is becoming a supported language of the ICPC world finals, so it'll be really useful to have it working in problemtools in advance of this year's regionals.

Code is  run using the kotlin command. File extensions:
 - *.kt (kotlin)
 - *.ktm (kotlin module)
 - *.kts (kotlin script)

### DIFF
--- a/problemtools/languages.yaml
+++ b/problemtools/languages.yaml
@@ -141,6 +141,13 @@ javascript:
     compile: '/usr/bin/js24 -c {files}'
     run: '/usr/bin/js24 "{mainfile}"'
 
+kotlin:
+    name: 'Kotlin'
+    priority: 250
+    files: '*.kt *.ktm *.kts'
+    compile: '/usr/bin/kotlinc -jvm-target 1.8 -d {path}/ -- {files}'
+    run: '/usr/bin/kotlin -J-XX:+UseSerialGC -J-Xss64m -J-Xms{memlim}m -J-Xmx{memlim}m -cp {path}/ {mainclass}'
+
 objectivec:
     name: 'Objective C'
     priority: 300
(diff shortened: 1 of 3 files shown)


problemtools/languages.yaml
```diff
@@ -141,6 +141,13 @@ javascript:
     compile: '/usr/bin/js24 -c {files}'
     run: '/usr/bin/js24 "{mainfile}"'
 
+kotlin:
+    name: 'Kotlin'
+    priority: 250
+    files: '*.kt *.ktm *.kts'
+    compile: '/usr/bin/kotlinc -jvm-target 1.8 {files} -include-runtime -d {binary}.jar'
+    run: '/usr/bin/kotlin -J-XX:+UseSerialGC -J-XX:ReservedCodeCacheSize=4m -J-XX:MaxMetaspaceSize=8m -J-XX:CompressedClassSpaceSize=8m -J-Xss5m -J-Xms5m -J-Xmx256m {binary}.jar'
```
meisterT: Can you explain what all the -J options that are not specified for java do (and why they're are relevant to Kotlin)?

Why do you include the full runtime if you do run kotlin later? In DOMjudge (https://github.com/DOMjudge/domjudge/blob/master/sql/files/defaultdata/kt/run) we just run kotlinc over all the sources and then run kotlinc with the same options as for Java (prefixed with -J of course).

hex539:
The -J options suppress Java trying to allocate a bunch of extra code cache
+ metaspace for the entire Kotlin standard library in the case that it's
compiled into the jar.

Point taken, the reason I did this originally is that this seems to be the
only way of convincing Kotlin to run a jar and take into account its entry
point instead of requiring MyClasName.kt to be passed in. I made a small,
rather hacky change to language.py instead to set mainclass properly.

Without the standard library included it's OK to just use the same runtime
args as java/scala.

hex539:
Selfishly closing since 2017 regionals are over and I'm not planning to write any model solutions in kotlin ever again.

PR #70, opened 2017-09-19
gimp development and aclocal tool

thuvh:
I created a vm with a clean Ubuntu server 16.04, then installed boost version 1.54.0 from source. After that, I use command

> pip install git+https://github.com/kattis/problemtools

There are two erros, they said that I need aclocal and libGIMP. After installing automake and libgimp-dev, the installation succeeded.


austrin:
@thuvh thanks for fix and sorry for taking a very very long time to look at it.  Since it was missing a few commas and in order to avoid all the merge commits, I made a separate commit to fix it.

PR #72, opened 2017-10-25
Add several tests on input validators.

### LINKED ISSUE #71: Input validation -- strict syntax checking
The input validation checker already warns if the input validator accepts things like the empty file, binary file, etc (https://github.com/Kattis/problemtools/blob/develop/problemtools/verifyproblem.py#L693).

When I am reviewing a problem for someone else I check for good input validators. I find that people often use inadequate syntax checking (e.g. their input validators check ranges on numbers, but not that each line is properly formatted with regular expressions). It's tedious for me to check for this and then explain why strict syntax checking is a good thing&trade;.

Here's an idea on how to automate this checking. Suppose we have verifyproblem play with the spacing of some input files that are in the problem already and pass validation. That is, add extra spaces and/or newlines where they already exist. If there is already an input format validator that uses strict syntax checking with regular expressions, then for many (most?) types of problems those input files should fail validation. If they don't then verifyproblem could give a warning.

If this is considered worthwhile, I can work on the changes.

ghamerly:
Fixes #71 

Check that modified input files don't validate, and warn if they do.

- add whitespace where it already exists
- add newlines where they already exist
- add random junk to the end of the input file
- add leading zeros to integers
- add trailing zeros to the decimal portion of real numbers

The downside is that this potentially checks every validator on every test input, which can be slow. But it's fast if the validators fail quickly (which is the correct behavior).

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -697,22 +697,16 @@ _JUNK_CASES = [
     ('a random text file with printable characters', ''.join(random.choice(string.printable) for _ in range(200))),
 ]
 
+def _build_junk_modifier(desc, pattern, repl):
+    p = re.compile(pattern)
+    return (desc, p.search, lambda text: p.sub(repl, text))
+
 _JUNK_MODIFICATIONS = [
-    ('add spaces where there already is whitespace',
-        lambda f: re.search(r'\s', f),
-        lambda f: re.sub(r'\s', lambda m: m.group(0) + ' ' * random.randint(0, 5), f)),
-    ('add newlines where there already are newlines',
-        lambda f: re.search('\n', f),
-        lambda f: re.sub('\n', lambda m: '\n' * random.randint(1, 5), f)),
-    ('add random junk to the end of the file',
-        lambda f: True,
-        lambda f: f + ''.join(random.choice(string.printable) for _ in range(200))),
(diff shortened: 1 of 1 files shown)


problemtools/verifyproblem.py
```diff
@@ -752,6 +769,30 @@ def collect_flags(group, flags):
                             break
                     else:
                         self.warning('No validator rejects %s with flags "%s"' % (desc, ' '.join(flags)))
+
+            def modified_input_validates(desc, applicable, modifier):
+                tested_any = False
+                for testcase in self._problem.testdata.get_all_testcases():
+                    with open(file_name, "wb") as f:
+                        infile = open(testcase.infile).read()
```
austrin: Do `with open(testcase.infile) as ....` to make sure it is closed.

problemtools/verifyproblem.py
```diff
@@ -752,6 +769,30 @@ def collect_flags(group, flags):
                             break
                     else:
                         self.warning('No validator rejects %s with flags "%s"' % (desc, ' '.join(flags)))
+
+            def modified_input_validates(desc, applicable, modifier):
```
austrin: `desc` is not used by the function.

ghamerly:
@austrin Thanks for the feedback; I have taken all your comments. See ea7b55a.

PR #73, opened 2017-10-27
Fix bug where we don't actually add more newlines

ghamerly:
This should have made it into #72 but I didn't notice the bug until after it was merged.


problemtools/verifyproblem.py
```diff
@@ -703,7 +703,7 @@ def _build_junk_modifier(desc, pattern, repl):
 
 _JUNK_MODIFICATIONS = [
     _build_junk_modifier('spaces added where there already is whitespace', r'\s', lambda m: m.group(0) + ' ' * random.randint(0, 5)),
```
austrin: Should this one also be updated (to `randint(1, 5)`)?

ghamerly: Yes, it should. Thanks. Will fix.

PR #77, opened 2018-04-10
Use validator output if it exits first, for interactive problems

simonlindholm:
As per the commit message:

>Before this change, if a validator detected, say, an invalid input
format and exited, it would close its output pipe, and the submission
would proceed to read EOF and possibly crash/TLE. The output of the
submission would then be displayed to the user.
>
>This isn't great -- it either results in confusing judgements, or
necessitates adding convoluted clean exit protocols to the problem to
deal with these cases. In the latter case, the contestant needs to add
probably-unnecessary hard-to-test code, and the description of the
protocol can bog down the problem statement (see e.g. "Go, Gopher" from
Google Code Jam 2018: https://goo.gl/ccT6t3).
>
>This commit takes the stand that if the validator exits first with a WA
status, that should take precedence. Note that this may result in WA
statuses where the time limit is exceeded.

With a few other cleanups thrown in.

Changes the output format of `interactive.cc`, I don't know if that matters for other consumers than kattis.

### DIFF
--- a/support/interactive/interactive.cc
+++ b/support/interactive/interactive.cc
@@ -15,51 +15,55 @@
 
 int report_fd, walltimelimit;
 
-int val_pid = -1, user_pid = -1;
-int user_status = -1, val_status = -1;
-static rusage user_ru;
-static rusage val_ru;
+volatile bool validator_first = false;
+volatile int val_pid = -1, user_pid = -1;
+volatile int user_status = -1, val_status = -1;
+volatile static rusage user_ru;
+volatile static rusage val_ru;
 
-double runtime(rusage *ru) {
+double runtime(volatile rusage *ru) {
 	if(ru == NULL) return 0;
 
 	struct timeval tv;
 	tv.tv_sec = ru->ru_utime.tv_sec + ru->ru_stime.tv_sec;
 	tv.tv_usec = ru->ru_utime.tv_usec + ru->ru_stime.tv_usec;
-	return tv.tv_sec + tv.tv_usec / 1000000.0;
+	return (double)tv.tv_sec + (double)tv.tv_usec / 1000000.0;
 }
 
 void report(int val_status, double val_time, int user_status, double user_time) {
 	FILE * fp = fdopen(report_fd, "w");
-	fprintf(fp, "%d %.6lf %d %.6lf", val_status, val_time, user_status, user_time);
+	fprintf(fp, "%d %.6lf %d %.6lf %s", val_status, val_time, user_status, user_time,
+			validator_first ? "validator" : "submission");
 	fclose(fp);
 }
 
-void walltime_handler(int a) {
+void walltime_handler(int) {
+	// TODO: make this race-free and signal safe. Right now there's a race
+	// between wait returning in main and the pid variables being set to -1,
+	// and we call non-signal safe functions...
+	// This is likely fine for problemtools, but not for real contest systems.
+	// The easiest way to fix this would probably be to kill(-1, SIGKILL) in
+	// this handler, set some sig_atomic_t variable, and let main() deal with
+	// the aftermath.
+
 	int u_stat = user_status, v_stat = val_status;
 	double u_time = 0;
 
 	// Check if validator has already quit while we were waiting for submission
-	if (val_pid != -1 && wait4(val_pid, &v_stat, WNOHANG, &val_ru) != val_pid)  {
+	if (val_pid != -1 && wait4(val_pid, &v_stat, WNOHANG, (rusage*)&val_ru) != val_pid)  {
 		kill(val_pid, SIGTERM);
-	} 
+	}
 
 	// Check submission resource usage and then kill it
-	if (user_pid != -1 && wait4(user_pid, &u_stat, WNOHANG, &user_ru) != user_pid)
+	if (user_pid != -1 && wait4(user_pid, &u_stat, WNOHANG, (rusage*)&user_ru) != user_pid) {
 		kill(user_pid, SIGKILL);
+	}
 	u_time = runtime(&user_ru);
 
 	if (u_stat == -1) {
-		// If validator already quit with WA but submission timed out
-		// on wall-time, don't tag submission as TLE but let validator
-		// decide.  (If validator quit with AC but submission kept
-		// running, tag submission as TLE.)
-		if (v_stat == (43 << 8)) u_stat = 0;
-		else {
-			u_stat = SIGUSR1;
-			u_time = walltimelimit;
-		}
-	} 
+		u_stat = SIGUSR1;
+		u_time = walltimelimit;
+	}
 
 	// If validator didn't yet give us something, assume WA
 	if (v_stat == -1) v_stat = 43 << 8;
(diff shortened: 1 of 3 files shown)


simonlindholm:
Now also with another (not yet tested) fix for validator SIGPIPEs. Unrelated but pretty similar to (and dependent upon) the rest of the PR.

simonlindholm:
@austrin Think you could review this? Does the approach make sense?

simonlindholm:
@austrin Ping! Or are you trying to safeguard yourself against interactive problems in the coding cup finals? ;)

simonlindholm:
@austrin Re-ping.

support/interactive/interactive.cc
```diff
@@ -15,56 +15,60 @@
...
 
-int val_pid = -1, user_pid = -1;
-int user_status = -1, val_status = -1;
-static rusage user_ru;
-static rusage val_ru;
+volatile bool validator_first = false;
+volatile int val_pid = -1, user_pid = -1;
+volatile int user_status = -1, val_status = -1;
+volatile static rusage user_ru;
+volatile static rusage val_ru;
```
austrin: Why do you think these (and others below) need to be volatile?

support/interactive/interactive.cc
```diff
@@ -236,32 +239,63 @@ int main(int argc, char **argv) {
...
+	close(fromuser[0]);
 
-	if(wait4(val_pid, &val_status, 0, &val_ru) == -1) {
-		perror("wait failed");
-		exit(1);
+	int remaining = 2;
+	while (remaining > 0) {
+		int status;
+		struct rusage ru;
+		int r = wait4(-1, &status, 0, &ru);
```
austrin: Change this to `wait3`?

support/interactive/interactive.cc
```diff
@@ -236,32 +239,63 @@ int main(int argc, char **argv) {
...
+			val_status = status;
+			memcpy((void*)&val_ru, &ru, sizeof(rusage));
+			val_pid = -1;
+			remaining--;
+			close(fromval[1]);
+		}
+		if (r == user_pid) {
+			// In case of broken pipes, let validator decide
+			user_status = (WIFSIGNALED(status) && WTERMSIG(status) == SIGPIPE ? 0 : status);
+			memcpy((void*)&val_ru, &ru, sizeof(rusage));
```
austrin: This should be `user_ru` not `val_ru`.

support/interactive/interactive.cc
simonlindholm: They are used from a signal handler, so it seemed "more correct" (even if still actually incorrect -- only `volatile sig_atomic_t`s and atomics are allowed, and even then we have race conditions that make the whole approach wrong). But until the signal handler is fixed properly, this may at least prevent the compiler from reordering some assignments etc. which I wouldn't be surprised if it does.

austrin: Makes sense!

simonlindholm:
> Change it so that if runtime > timelim then result is TLE regardless of what validator says (but if submission crashed or exceeded walltime limit, which are probably more common ways of failing after validator has terminated, then validator gets its say).

With the PR as is walltime limit exceeded will practically never happen when the validator exits early, because then we close all pipes and you cannot block on them forever. We could of course decide not to close the pipes and then wait for the wall time limit to hit, but that seems somewhat silly in that it only prolongs time until error...

How about setting runtime = min(validator exit time, submission exit time)?

austrin:
> With the PR as is walltime limit exceeded will practically never happen when the validator exits early, because then we close all pipes and you cannot block on them forever. We could of course decide not to close the pipes and then wait for the wall time limit to hit, but that seems somewhat silly in that it only prolongs time until error...

Right, I agree.  (But the crashing part is still probably the most common way for a solution to fail after the validator has terminated.)

> How about setting runtime = min(validator exit time, submission exit time)?

I don't think that makes sense, the two programs may need very differing amounts of CPU time and there is no reason why the CPU time of the validator would be any reasonable measure of the CPU time of the submission.  However, one possibility that I think would make a lot of sense would be to fetch the current CPU time of the submission process at the time the validator exits.

simonlindholm:
> Right, I agree. (But the crashing part is still probably the most common way for a solution to fail after the validator has terminated.)

If it's just "the most common way" then it doesn't help because you still need to know about the edge case... Also, I'm not sure my experiences agree with crashing being more likely than TLE.

> However, one possibility that I think would make a lot of sense would be to fetch the current CPU time of the submission process at the time the validator exits.

Sounds good! I'll see if I can make this work here and in Kattis (tomorrow or so).

simonlindholm:
Also worth mentioning: the first approach I had to this was to kill the submission when the validator exited; however, it turns out that that doesn't work well with isolate, which runs as `sudo` and is thus unkillable without complex hackery. Fetching the current CPU time should have the same effect as that (UI-wise), nicely enough.

niemela:
>> However, one possibility that I think would make a lot of sense would be to fetch the current CPU time of the submission process at the time the validator exits.
> 
> Sounds good! I'll see if I can make this work here and in Kattis (tomorrow or so).

Agreed, that would give the expected results. Because what you would *actually* really want is...

> Also worth mentioning: the first approach I had to this was to kill the submission when the validator exited; 

... to kill the submissions as soon as the validator has decided to give up.

> however, it turns out that that doesn't work well with isolate, which runs as sudo and is thus unkillable without complex hackery. Fetching the current CPU time should have the same effect as that (UI-wise), nicely enough.

👍

simonlindholm:
It appears to be pretty hard to get to that information. `wait4` doesn't return usage information about running processes; you have to parse `/proc/<id>/stat`. However, that only gives you the CPU usage of the process and the sum of CPU usage of terminated and waited-for children, and if the process is, say, `isolate`, both of these values are zero. Which means we only have bad options remaining:

* do some sort of process tree traversal, with all the racefulness that implies. Could be improved by SIGSTOP'ing the process first, but ugh. In a sense it's OK if we mess up and return an invalid value because the result will be WA anyway, but...
* kill the submission as soon as the validator stops, and on Kattis's side try to run the program with extended caps that allow killing privileged processes (`setcap cap_kill+ep`). Hard to get right (security issues, and is it even OK to kill isolate half-way through?).
* pull a runtime out of thin air, e.g. 0 or exactly the TL.

austrin:
There is also the (also bad) option of keeping TLE if process uses too much CPU.  I would slightly prefer that, but I am also fine with capping runtime at timelim in the case when validator decides on WA before the user process exits.

simonlindholm:
> There is also the (also bad) option of keeping TLE if process uses too much CPU. I would slightly prefer that, but I am also fine with capping runtime at timelim in the case when validator decides on WA before the user process exits.

In my view that's more or less equivalent to (maybe slightly worse than) wontfixing this PR, since it would still require explaining in the problem statement that you must (somehow, by some problem-specific scheme) exit nicely when the validator exits to avoid confusing verdicts. But sure, it's an option.

Anyway, I pushed a commit for the `runtime = min(runtime, timelim)` behavior (not very well tested but I expect it to work), and another to address the review comments you had earlier.

austrin:
Sounds good. How about also adding a test submission to `examples/guess` which gets WA this way?

simonlindholm:
Thanks for merging! I agree that a test is a great idea, I'll try to get to that at some point if you don't do it before me. Also, does this need some sort of spec update or coordination with other judges?

PR #80, opened 2018-06-16
Support test case reuse using symbolic links

simonlindholm:
@niemela and I discussed this, and concluded that this was probably the most reasonable way to support test case reuse between different test groups. (Basing it on hashes would also be an option but is less explicit and makes archiving more expensive. Using graders to simulate test-case reuse seemed like a good idea at first but turns out not to work because test group results can be displayed in the contest system UI.)

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -32,7 +32,6 @@ def is_TLE(status, may_signal_with_usr1=False):
 def is_RTE(status):
     return not os.WIFEXITED(status) or os.WEXITSTATUS(status)
 
-
 class SubmissionResult:
     def __init__(self, verdict, score=None, testcase=None, reason=None):
         self.verdict = verdict
@@ -44,6 +43,11 @@ class SubmissionResult:
         self.ac_runtime = -1.0
         self.ac_runtime_testcase = None
         self.validator_first = False
+
+    def set_ac_runtime(self):
+        if self.verdict == 'AC':
+            self.ac_runtime = self.runtime
+            self.ac_runtime_testcase = self.runtime_testcase
 
     def __str__(self):
         verdict = self.verdict
@@ -106,6 +110,9 @@ class TestCase(ProblemAspect):
         self.ansfile = base + '.ans'
         self._problem = problem
         self.testcasegroup = testcasegroup
+        self.reuse_result_from = None
+        self._result_cache = (None, None)
+        problem.testcase_by_infile[self.infile] = self
 
     def check_newlines(self, filename):
         with open(filename, 'r') as f:
@@ -139,6 +146,7 @@ class TestCase(ProblemAspect):
                     self.error('judge answer file got %s' % val_res)
                 else:
                     self.warning('judge answer file got %s' % val_res)
+        self._check_symlinks()
         return self._check_res
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


simonlindholm:
Right, argument against file hashes: then we don't know whether to warn about duplicate test cases. We could silence that warning iff the cases are in different groups, but it would reduce usefulness of the warning slightly.

austrin:
In principle this sounds like a nice feature though I am somewhat apprehensive about the additional code complexity.  One half-baked thought: would it (a) be sufficient (for your purposes), and (b) less complex, to instead only allow symlinks between groups, rather than individual test files?

simonlindholm:
So that e.g. `secret/group5/group1` is a symlink to `secret/group1`? That *might* be sufficient for us, but it's not optimal:
- we currently support reusing individual test cases, so in that case we would need to copy files and live with the warning
- the semantics would have to be to include all files in the symlinked group, not the group's result, because the group has a score which may be incorrect

I think the latter point also means that this isn't any less complex to implement, and it loses the nice property that a judge that doesn't special-case symlinks continues to work.

problemtools/verifyproblem.py
```diff
@@ -145,7 +154,50 @@ def __str__(self):
...
+            return True
+        nicepath = os.path.relpath(self.infile, self._problem.probdir)
+        in_target = os.path.realpath(self.infile)
+        ans_target = os.path.realpath(self.ansfile)
+        if not in_target.endswith('.in'):
+            self.error("Symbolic link does not point to a .in file for input '%s'" % nicepath)
+            return False
+        if ans_target != in_target[:-3] + '.ans':
+            self.error("Symbolic link '%s' must have a corresponding link for answer file" % nicepath)
+            return False
```
austrin: Is it conceivable to ever have a situation where one would want to supply a different answer file for the duplicated case?  E.g. one could consider an optimization problem where one gets AC if one beats the answer file produced by some reference solution, and the same case appears several times but with different thresholds of what to beat.  But even if that happens I guess it is so rare that it is not worth the extra complexity, so then one should just duplicate the input file.

problemtools/verifyproblem.py
```diff
@@ -1,5 +1,6 @@
 #! /usr/bin/env python2
 # -*- coding: utf-8 -*-
+from __future__ import print_function
```
austrin: Could you remove the python3 print stuff from the PR?  (Not saying it is a bad idea, but better then to make a separate PR which applies this change to all of problemtools, not just to verifyproblem.py.)

problemtools/verifyproblem.py
```diff
@@ -145,7 +154,50 @@ def __str__(self):
     def matches_filter(self, filter_re):
         return filter_re.search(self.strip_path_prefix(self._base)) is not None
 
+    def set_symlinks(self):
+        if not os.path.islink(self.infile):
+            return
```
austrin: Aren't these two lines redundant given the check below that `target != self.infile`?

problemtools/verifyproblem.py
```diff
@@ -170,18 +222,24 @@ def run_submission(self, sub, args, timelim_low=1000, timelim_high=1000):
...
-            res1.ac_runtime_testcase = res1.runtime_testcase
-        if res2.verdict == 'AC':
-            res2.ac_runtime = res2.runtime
-            res2.ac_runtime_testcase = res2.runtime_testcase
-        self.info('Test file result: %s)' % (res1))
-        return (res1, res2)
+        self._result_cache = (cache_key, (res1, res2))
+        return (res1, res2, False)
+
+    def _apply_default_score(self, res):
```
austrin: This function is a bit misnamed since it does more than just applying default score.  Maybe `_init_result_defaults`?

problemtools/verifyproblem.py
simonlindholm: In that case output validator flags would need to differ, and we emit an error if you use symlinks (precisely because the judge can't just reuse results). One should indeed duplicate the input file in that case.

problemtools/verifyproblem.py
simonlindholm: Why the opposition to drive-by cleanups :( But sure, I can split that out...

austrin: In this case it was mostly an opposition to drive-by cleanups that leave things inconsistent (with some programs using Python 2 prints and others using Python 3 prints). (In other cases it has been a case of "let's extract the parts that can be merged immediately to reduce rot.")

problemtools/verifyproblem.py
simonlindholm: They are, but I think they add clarity that this only affects symlinks. OTOH, I guess the `target != self.infile` check is redundant since a symlink pointing to itself would create an unresolvable infinite loop.

problemtools/verifyproblem.py
austrin: No in the hypothetical example the output validator would do the same thing in both cases (check that output solution is at least as good as answer file solution), only the answer files would differ.  But we anyway agree that this is not a scenario that should be covered so this doesn't matter.

simonlindholm:
> What happens if the ans file is a symlink but not the in file? I guess the answer is "the same as today", and maybe that's fine?

Yeah, same as today. There's a lot of places where we don't handle symlinks and that's fine, since it's harmless and never happens in practice anyway. The reason for the current setup was basically, we need some signal that reuse is wanted, and ".in file is a symlink" seemed like a good one. Then we need to assert that reuse is actually possible. I could make the condition "either .in or .ans file is a symlink", though, if wanted.

niemela:
@simonlindholm Please rebase and I will merge.

PR #81, opened 2018-06-24
Add a "partially accepted" directory

simonlindholm:
It works like "accepted", except that:

- it is an error if a submission in `partially_accepted` gets full score
- it is a warning if a submission in `accepted` does not get full score
- submissions in `partially_accepted` do not affect the time limit (so this gives us a way to test TLE solutions for subtasks).

As a bonus, this allows us to give a warning whenever a submission in `accepted` doesn't get AC on the sample group. This is important since if the sample group is configured to give 0 points -- as it usually is for scoring problems -- verifyproblem won't otherwise give any hints about wrongly set sample answers. (We were very close to making this mistake for the Swedish Coding Cup.)

We may still want to add "expected score" annotations to submissions in the future; this does not preclude that.

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -1216,12 +1216,14 @@ class OutputValidators(ProblemAspect):
 
 class Submissions(ProblemAspect):
     _SUB_REGEXP = re.compile(r'^[a-zA-Z0-9][a-zA-Z0-9_.-]*[a-zA-Z0-9](\.c\+\+)?$')
+    # (verdict, directory, required)
     _VERDICTS = [
         ['AC', 'accepted', True],
+        ['PAC', 'partially_accepted', False],
         ['WA', 'wrong_answer', False],
         ['RTE', 'run_time_error', False],
         ['TLE', 'time_limit_exceeded', False],
-        ]
+    ]
 
     def __init__(self, problem):
         self._submissions = {}
@@ -1239,22 +1241,37 @@ class Submissions(ProblemAspect):
     def __str__(self):
         return 'submissions'
 
-    def check_submission(self, sub, args, expected_verdict, timelim_low, timelim_high):
+    def check_submission(self, sub, args, expected_verdict, timelim, timelim_low, timelim_high):
         desc = '%s submission %s' % (expected_verdict, sub)
-        (result1, result2) = self._problem.testdata.run_submission(sub, args, timelim_low, timelim_high)
+        partial = False
+        if expected_verdict == 'PAC':
+            # For partially accepted solutions, use the low timelim instead of the real one,
+            # to make sure we have margin in both directions.
+            expected_verdict = 'AC'
+            partial = True
+            timelim = timelim_low
 
-        if result1.verdict != result2.verdict:
-            r1, r2 = result1.verdict, result2.verdict
-            self.warning('%s sensitive to time limit: limit of %s secs -> %s, limit of %s secs -> %s' % (desc, timelim_low, r1, timelim_high, r2))
+        (result1, result2) = self._problem.testdata.run_submission(sub, args, timelim, timelim_high)
 
-        if result1.verdict == expected_verdict:
+        if result1.verdict != result2.verdict or result1.score != result2.score:
+            r1, r2 = result1, result2 if result1.verdict == result2.verdict else result1.verdict, result2.verdict
+            self.warning('%s sensitive to time limit: limit of %s secs -> %s, limit of %s secs -> %s' % (desc, timelim, r1, timelim_high, r2))
+
+        if partial and self.fully_accepted(result1):
+            self.warning('%s got %s' % (desc, result1))
+        elif result1.verdict == expected_verdict:
             self.msg('   %s OK: %s' % (desc, result1))
-        elif result2.verdict == expected_verdict:
+        elif result2.verdict == expected_verdict and not (partial and self.fully_accepted(result2)):
             self.msg('   %s OK with extra time: %s' % (desc, result2))
         else:
             self.error('%s got %s' % (desc, result1), result2.additional_info)
 
         return result1
+
+    def fully_accepted(self, result):
+        min_score, max_score = self._problem.testdata.get_score_range()
+        best_score = min_score if self._problem.config.get('grading')['objective'] == 'min' else max_score
+        return result.verdict == 'AC' and (not self._problem.is_scoring or result.score == best_score)
 
     def check(self, args):
         if self._check_res is not None:
@@ -1262,14 +1279,18 @@ class Submissions(ProblemAspect):
         self._check_res = True
 
         limits = self._problem.config.get('limits')
+        time_multiplier = limits['time_multiplier']
+        safety_margin = limits['time_safety_margin']
 
-        timelim_margin = 300  # 5 minutes
+        timelim_margin_lo = 300  # 5 minutes
+        timelim_margin = 300
         timelim = 300
+
         if 'time_for_AC_submissions' in limits:
             timelim = timelim_margin = limits['time_for_AC_submissions']
         if args.fixed_timelim is not None:
             timelim = args.fixed_timelim
-            timelim_margin = timelim * limits['time_safety_margin']
+            timelim_margin = int(round(timelim * safety_margin))
 
         for verdict in Submissions._VERDICTS:
             acr = verdict[0]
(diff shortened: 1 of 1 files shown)


austrin:
Before adding something like this, the problem format spec needs to be updated accordingly.  It also seems the PR contains several unrelated cleanup changes.

simonlindholm:
Sure. Can I can go ahead and update it, or do we need buy-in from someone else?

Do you want me to split out the cleanups into separate PRs? I figured they were harmless.

simonlindholm:
> I could change this to a warning instead of an error if we don't want to make this breaking change to the problem format.

(Changed; on further thought it seemed unnecessary to introduce this breaking change.)

simonlindholm:
Also, @niemela: you seemed to warm up to this change when we talked it over. Are you OK with taking it in its current state and adding it to the problem format?

simonlindholm:
What is this blocked on, btw? Can we speed it up? This is the single largest problemtools pain-point we have for PO.

simonlindholm:
@niemela can we get this merged, pretty please? As mentioned, it's a large pain point for PO, and I think the design is the right one even given score annotations.

austrin:
Once the problem format spec is updated to include this feature, I'm happy to take a look at these changes.  (But they can not be merged until the spec is updated.)

simonlindholm:
@niemela can we get this added to the spec? The changes are as follows:

- for scoring problems, submissions in accepted/ give a warning if they don't get the best score (I don't think this needs to be in the spec)
- there is a partially_accepted/ directory added, with the following semantics:
  * pass-fail problems must not have any entries in this directory
  * solutions in partially_accepted/ must get Accepted
  * it is *not* currently a warning if a submission in partially_accepted/ gets the best score; that could be added later without a spec change, but it seems harmless (since you already see the scores printed) and it's useful for slow solutions at the moment
  * time limits are not set based on partially_accepted/ solutions
  * time limit safety margin is checked in both directions for partially_accepted/ solutions. Concretely, we set `timelim_margin_lo = max(1, min(round(exact_timelim / safety_margin), timelim - 1))`, `timelim_margin_hi = max(timelim + 1, round(exact_timelim * safety_margin))`, and if the solution does not give the same score given both these time limits, that results in a warning.

It can be argued that some of these behaviors are orthogonal and should be separately configurable. In practice, however, this is basically the behavior you want when working with problems with subgroups. Optionally you might want more control over expected scores, but it works reasonably well to use a naming scheme and manual checks for that, and in the future it's extensible using yaml files or source code annotations. The current plan for the future is to make it possible for the user to add additional directories with yaml files which describe behaviors of the group, such as expected verdict(s), whether it's used for setting time limit, what the expected score is, ...

(cc @SuprDewd, who also wants this for Iceland's olympiad)

simonlindholm:
> it is not currently a warning if a submission in partially_accepted/ gets the best score; that could be added later without a spec change, but it seems harmless (since you already see the scores printed) and it's useful for slow solutions at the moment

hm, I misread my PR, it's currently an error. Let's go with error, I guess; it's backwards compatible to change later.

niemela:
I think this sounds good. I'll try to update the spec.

I think we will probably want to introduce a lower safety margin eventually, but it could reasonably default to `1/safety margin` so I think we should start with this and introduce it only if it turns out we really need it.

austrin:
@niemela please ping me when spec is updated and I'll take a look (both at the spec updates and PR).

niemela:
@simonlindholm I have added this to the spec.

(I noticed that the verification description is *very* out of date. For now I only added a comment at the top, but we should either fix this or remove it.)

niemela:
@austrin ping.

ghamerly:
@niemela 

Some folks are being confused by the current specification at https://www.problemarchive.org/wiki/index.php/Problem_Format#Included_Code
because it appears there as if `partially_accepted` is currently supported. Shouldn't that page show it highlighted in yellow?

I would change the spec, but I cannot locate my wiki account at the moment.

niemela:
@ghamerly I have fixed the spec, sorry about missing that to begin with.

That said, another (better) fix would be to merge this PR :)

problemtools/verifyproblem.py
```diff
@@ -1221,35 +1232,60 @@ def __init__(self, problem):
...
-            self.warning('%s sensitive to time limit: limit of %s secs -> %s, limit of %s secs -> %s' % (desc, timelim_low, r1, timelim_high, r2))
-
-        if result1.verdict == expected_verdict:
+        partial = False
+        if expected_verdict == 'PAC':
+            # For partially accepted solutions, use the low timelim instead of the real one,
+            # to make sure we have margin in both directions.
+            expected_verdict = 'AC'
+            partial = True
+            timelim = timelim_low
```
austrin: As far as I understand, the purpose of this is to give a warning if, when running the submission with half the time limit (where "half"=whatever the 1/safety margin is, 0.5 by default), it gives a different result than when running with twice the time limit (where "twice"=...)?  Why is this necessary, and what makes PAC different from other verdicts that it is warranted for PAC but not for other verdicts?

problemtools/verifyproblem.py
```diff
@@ -1221,35 +1232,60 @@ def __init__(self, problem):
...
             self.msg('   %s OK with extra time: %s' % (desc, result2))
         else:
             self.error('%s got %s' % (desc, result1))
 
         return result1
 
+    def fully_accepted(self, result):
+        min_score, max_score = self._problem.testdata.get_score_range()
+        best_score = min_score if self._problem.config.get('grading')['objective'] == 'min' else max_score
+        return result.verdict == 'AC' and (not self._problem.is_scoring or result.score == best_score)
```
austrin: I feel that this warning is going to be annoying for open-ended heuristic problems without a clear upper bound on the optimum.  So I suggest not giving the warning if max_score/min_score (whichever is applicable) is inf/-inf.

problemtools/verifyproblem.py
```diff
@@ -182,12 +185,15 @@ def _check_symlinks(self):
             return False
         return True
 
-    def run_submission(self, sub, args, timelim_low, timelim_high):
+    def run_submission(self, sub, args, timelim_low, timelim_high, subinfo):
```
austrin: I don't like introducing this dict of "whatever extra data we realize we want to pass around".  In this case it is not even really the case of info that is being passed to `run_submission`, but rather a mechanism for `run_submission` to send additional result data back.

I would prefer adding a "failed_samples" field to the SubmissionResult class and passing it back upwards that way.  (The SubmissionResult class is already suffering from being a bit of a "let's add everything that occurs to us here" mess, but at least then that messiness remains contained to that class and we're not adding another mechanism with the same issues.)

austrin:
> (I noticed that the verification description is very out of date. For now I only added a comment at the top, but we should either fix this or remove it.)

@niemela I did not see any comment about this, maybe you forgot to save?

problemtools/verifyproblem.py
simonlindholm: The idea is to make sure that if you include a solution in partially_accepted/, making it slightly slower or slightly faster should not affect its score.

For e.g. a TLE solution, we check that making it slightly faster doesn't make it get accepted, but checking that making it slower doesn't affect its score is pointless since it scores 0.

Arguably you'd want to be able to say "I wrote a slow Python solution, please ensure its score doesn't change when making it 1.5 times slower or 20 times faster", but we're going with only time_safety_margin for now in the hope that that's good enough in practice.

problemtools/verifyproblem.py
simonlindholm: Good point, will do

problemtools/verifyproblem.py
simonlindholm: Makes sense. I'll look into this and see if it's doable.

simonlindholm:
> @niemela I did not see any comment about this, maybe you forgot to save?

https://www.problemarchive.org/wiki/index.php/Problem_Format_Verification

problemtools/verifyproblem.py
niemela: To clarify (not sure if it's needed). We believe that for completeness you would actually need to specify an upper as well as a lower safety margin. It's not (at all) clear to me that these must always be the same. OTOH, we suspect that in practice having the lower be the same (or the inverse, depending on how you look at it) as the upper (i.e. the only one we have) is good enough.

Also, if it turns out that you *do* need to specify a lower margin once in a while, then what we are using now is certainly a very reasonable default, so it will be safe to introduce it later.

simonlindholm:
Updated with the suggested comments. Also turned full-score partially-accepted back into a warning because it seems strictly more useful.

problemtools/verifyproblem.py
```diff
@@ -1239,37 +1252,72 @@ def __init__(self, problem):
...
+                # For some heuristic problems, this is expected. Thus, only warn.
+                self.warning('%s did not attain full score (consider moving it to partially_accepted)' % desc)
+        elif result2.verdict == expected_verdict and not (partial and self.fully_accepted(result2)):
             self.msg('   %s OK with extra time: %s' % (desc, result2))
         else:
             self.error('%s got %s' % (desc, result1), result2.additional_info)
 
         return result1
 
+    def full_score_attainable(self):
```
austrin: Nitpick: I would call this `full_score_finite` instead, because it is more immediately obvious what it means and more accurate (the name `attainable` hints that it somehow tries to solve the undecidable problem of figuring out which scores the grader might produce).

austrin:
Not super-sold on the two-sided safety margin, but let's roll with it.

@simonlindholm can you fix the last nitpick I just added and then merge?

simonlindholm:
Updated.

niemela:
@simonlindholm Done.

Please confirm that this was the correct change:
https://www.problemarchive.org/wiki/index.php?title=Problem_Format&type=revision&diff=395&oldid=393

simonlindholm:
The description of `first_error` shouldn't be yellow, and it has a stray `</s>` at `first_error</s>` (speaking of which, there's also a `<s>` with missing `</s>` in `<s>Generators`). Looks good other than that.

PR #83, opened 2018-06-24
Add subgroups visibility field, and require that visible subgroups have numerical names

simonlindholm:
For discussion. As I see it, we need either a visibility field, or to say that subgroups should always be displayed in the judge UI (but we have a bit of legacy content for which that doesn't make sense). Certainly I think the way Kattis does it right now, where subgroup visibilities are configured per-site within the judge, is broken. Subgroup visibility is a property of the problem: either the problem statement mentions what subgroups exist, in which case the judge should always show group results, or it doesn't, and it shouldn't.

(Also, we have had problems with the current setup -- one contest was accidentally misconfigured and we heard only after the contest that contestants had gotten no feedback about test groups... Putting the config flag in testdata.yaml eliminates that problem.)

This PR adds a subgroup visibility field, and defaults it to off. This gives us the unique chance to make a (non-breaking!) change that visible groups have to have (mostly-)numerical names, which a judge can format as `Group N` in any language. The current situation is that subgroup presentations are left unspecified -- Kattis shows raw directory names which is not good from a localization or UI control perspective.

The topmost `subgroups` flag is intended to be ignored (other than for inheritance into subconfigs); the judge should itself decide whether to show sample and secret group outcomes.

Possible points of discussion (though I'm quite happy with this PR as is):
* what should the default visibility be?
* what should the flag be called?
* should the flag affect whether all the children are visible, or whether the current group is? (IMO the first is simpler, and having only a subset of subgroups show up is confusing.) Or should it even be in problem.yaml and affect all groups? (Simpler yet, but loses too much flexibility I suspect.)
* do we want to impose this requirement on group names?
* what precise pattern do we accept? I felt that `5.2.4a` was reasonable subgroup name, so wrote a regex that accepts that... but we could just as well limit ourselves to purely numerical names, or to some other pattern.
* should group names be a function of directory names, or of some testdata.yaml flag, or auto-generated (group 1, group 2, ...)? (IMO putting them into testdata.yaml isn't a win, and auto-generating is too risky... but it's worth noting the option)

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -198,8 +198,9 @@ class TestCaseGroup(ProblemAspect):
                        'on_reject': 'break',
                        'accept_score': 1.0,
                        'reject_score': 0.0,
-                       'range': '-inf +inf'
-   }
+                       'range': '-inf +inf',
+                       'subgroups': 'hidden'
+    }
 
     _SCORING_ONLY_KEYS = ['accept_score', 'reject_score', 'range']
 
@@ -306,8 +307,11 @@ class TestCaseGroup(ProblemAspect):
                 if self.config.get(key) is not None:
                     self.error("Key '%s' is only applicable for scoring problems, this is a pass-fail problem" % key)
 
-        if not self.config['on_reject'] in ['break', 'continue']:
+        if self.config['on_reject'] not in ['break', 'continue']:
             self.error("Invalid value '%s' for on_reject policy" % self.config['on_reject'])
+
+        if self.config['subgroups'] not in ['visible', 'hidden']:
+            self.error("Invalid value '%s' for subgroups visibility" % self.config['subgroups'])
 
         if self._problem.config.get('type') == 'scoring':
             # Check grading
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -209,6 +209,8 @@ class TestCaseGroup(ProblemAspect):
         self._parent = parent
         self._problem = problem
         self._datadir = datadir
+        self._full_name = os.path.relpath(self._datadir, os.path.join(self._problem.probdir))
+        self._name = os.path.basename(self._datadir)
         self.debug('  Loading test data group %s' % datadir)
         configfile = os.path.join(self._datadir, 'testdata.yaml')
         if os.path.isfile(configfile):
@@ -260,7 +262,7 @@ class TestCaseGroup(ProblemAspect):
 
 
     def __str__(self):
-        return 'test case group %s' % os.path.relpath(self._datadir, os.path.join(self._problem.probdir))
+        return 'test case group %s' % self._full_name
 
 
     def matches_filter(self, filter_re):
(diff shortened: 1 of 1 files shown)


niemela:
A few comments to start the discussion?

> subgroups should always be displayed in the judge UI (but we have a bit of legacy content for which that doesn't make sense)

Could you elaborate on this?

> Subgroup visibility is a property of the problem

Agreed. (And form this the rest of your first paragraph follows).

> visible groups have to have (mostly-)numerical names

Specifically they must start with a dot-separated list of "number strings" (in lack of a better name for strings of digits) optionally followed by a single character? I don't like this.

> The current situation is that subgroup presentations are left unspecified

No, the specification states that the order is lexicographical order on the path. It also suggests that numbered prefixes can be used to get any desired order. I don't think that this needs changing.

> what should the default visibility be?

Visible.

> what should the flag be called?

"visible"?

> should the flag affect whether all the children are visible, or whether the current group is? (IMO the first is simpler, and having only a subset of subgroups show up is confusing.) Or should it even be in problem.yaml and affect all groups? (Simpler yet, but loses too much flexibility I suspect.)

Having a group be invisible but some of it's children visible does not make much sense, does it?

Is there a use case forever having subgroups (below the first level) visible?

Is there a use case for having groups on the first level invisible?

The original intent was that all the first level groups should be visible and nothing else. I.e. what you get if the answers to my questions above are all "no".

> do we want to impose this requirement on group names?

No!

> what precise pattern do we accept? I felt that 5.2.4a was reasonable subgroup name, so wrote a regex that accepts that... but we could just as well limit ourselves to purely numerical names, or to some other pattern.

If we go this route (which I don't think we should) than this is a reasonable pattern.

> should group names be a function of directory names, or of some testdata.yaml flag, or auto-generated (group 1, group 2, ...)? (IMO putting them into testdata.yaml isn't a win, and auto-generating is too risky... but it's worth noting the option)

If groups needs names I think they should go in testdata.yaml. Could you explain why that is not good?

simonlindholm:
> @simonlindholm Did I miss something?

Not really. We discussed what the default for the setting should be but didn't reach a conclusion. I'll close this and make a new PR with the other behavior.

austrin:
@niemela can you update the problem format specification accordingly, and I will then check the next PR against that?

simonlindholm:
So there are a few open questions:

- are we OK with the decision to put the flag in problem.yaml, having it provide only a single layer of groups with auto-generated names, and only for scoring problems? (I'm pretty comfortable with that. It's extensible towards explicit group names in the future in the form of testdata.yaml flags, and we could also later add, say, `testgroups: recursive`, to handle another 90% of the 0.1% of cases for which a single layer isn't enough.)
- what's the flag name? (I can't really think of a good one... `testgroups: yes|no`?)
- what the default? (I don't know... Doesn't matter hugely, we cargo-cult these files anyway.)

Do we need to spec the sort order for warnings?

FWIW, I've been quite happy with sending PRs before updating the spec, because I feel like it makes me think, gets me a better sense of the exact details and interactions involved, provokes discussion, and the code is simple enough to change when the spec has been decided.

austrin:
How about something like `testgroup_visbility_level` which is an integer (0 being what you call "no", 1 being what you call "yes", a large number being what you call `recursive`, and in general x meaning that test groups up to depth x are visible)?  Not sure whether I like it or not, just an idea.

For backwards compatibility I would prefer the default to be 0/no.

simonlindholm:
Certainly a possibility, though it's rather verbose and draws focus to the uncommon case.

simonlindholm:
Opened #84 with a proof-of-concept implementation.

PR #84, opened 2018-07-04
Add subgroups visibility field, take 2

simonlindholm:
Replaces #83. This PR is mostly just a proof-of-concept for the idea of warning for confusing sort order. It's not ready to merge; still waiting on @niemela to comment on what flag name/default to use, and to update the spec.

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -446,6 +446,39 @@ class TestCaseGroup(ProblemAspect):
             if not f[:-4] + '.in' in infiles:
                 self.error("No matching input file for answer '%s'" % f)
 
+        # Check whether a <= b according to a natural sorting where numeric components
+        # are compactified, so that e.g. "a" < "a1" < "a2" < "a10" = "a010" < "a10a".
+        def natural_sort_le(a, b):
+            a += '\0'
+            b += '\0'
+            i = j = 0
+            def parse_num(s, i):
+                ret = 0
+                while ord('0') <= ord(s[i]) <= ord('9'):
+                    ret = ret * 10 + ord(s[i]) - ord('0')
+                    i += 1
+                return ret, i
+            while i < len(a) and j < len(b):
+                if ord('0') <= ord(a[i]) <= ord('9') and ord('0') <= ord(b[i]) <= ord('9'):
+                    anum,i = parse_num(a, i)
+                    bnum,j = parse_num(b, j)
+                    if anum == bnum:
+                        continue
+                    return anum < bnum
+                if a[i] == b[j]:
+                    i += 1
+                    j += 1
+                    continue
+                return a[i] < b[j]
+            return True
+
+        last_testgroup_name = ''
+        for group in self.get_subgroups():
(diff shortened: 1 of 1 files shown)


problemtools/verifyproblem.py
```diff
@@ -549,6 +583,11 @@ def check(self, args):
         elif self._data['license'] == 'unknown':
             self.warning("License is 'unknown'")
 
+        if self._data['grading']['testgroups'] not in [True, False]:
```
niemela: Change key name to `test_data_groups`.

problemtools/verifyproblem.py
```diff
@@ -549,6 +583,11 @@ def check(self, args):
...
+            self.error("Invalid value for grading.testgroups: %s" % self._data['grading']['testgroups'])
+        elif self._data['grading']['testgroups'] and self._data['type'] == 'pass-fail':
+            self.error("Test groups are only supported for scoring problems, this is a pass-fail problem")
```
niemela: The `grading` map can only be used with scoring problems, so this check should be one level higher.

simonlindholm:
> Please comment on the name chosen (test_data_groups).

I'd prefer `testgroups`, but it's not a big deal. (Reasoning: either way it's confusing -- the flag doesn't turn on test data groups but merely displays them -- but I can't come up with a much better term. "testgroups" is a bit more informal, familiar and sounds better, and perhaps ties a bit closer to the UI concept than to the spec's "test data groups". Not that people care about formal spec terms. Which might also reduce the reason for using the spec term, in addition to it sounding odd.)

problemtools/verifyproblem.py
niemela: Ok, fair enough...

niemela:
> I'd prefer `testgroups`, but it's not a big deal. (Reasoning: either way it's confusing -- the flag doesn't turn on test data groups but merely displays them -- but I can't come up with a much better term. "testgroups" is a bit more informal, familiar and sounds better, and perhaps ties a bit closer to the UI concept than to the spec's "test data groups". Not that people care about formal spec terms. Which might also reduce the reason for using the spec term, in addition to it sounding odd.)

If the term ("Test Data Group") is not good we should update the spec, not just use different terms in different places. I *do* think that "Test Data Group" is better than "Test Group".

Regarding spaces or no spaces, there is a space in "Test Group" so there should be one in the key as well.

I still think `test_data_group` is better than `testgroup` (or `test_group`).

simonlindholm:
Merged with that name change. Please update the wiki @niemela.

simonlindholm:
> Please update the wiki @niemela.

Ping. (Not urgent, but we'll forget this at some point...)

simonlindholm:
Ping @niemela...

austrin:
@simonlindholm can you summarize what are the spec updates to be made and I can make them?

simonlindholm:
`test_data_groups` should change to `show_test_data_groups`.

austrin:
Ah ok it was that trivial.  Fixed now.

simonlindholm:
Thanks!

Issue #51, opened 2016-10-14
Support 'wrong' as test data category


meisterT:
It would be nice if we could add test data to verify that the input checker fails on them but they should not be used for real judging.

meisterT:
It would be also nice to have a category for output validator checking.

austrin:
Agree that this would be nice.  Do you have concrete suggestions/preferences for how to extend the problem format to include this?

For input files, one could consider adding something like `input_format_validators/bad_inputs` containing .in files, or `data/bad_inputs`.

For output files, maybe one wants to be able to add incorrect outputs to some of the actual test cases, rather than adding new cases?  If so, maybe allowing adding data files with suffix `.out.wrong-<id>`?

@niemela do you want to weigh in on this?

meisterT:
For separation of concerns input_format_validators/bad_inputs is probably nicer. Your suggestion for output files sounds good to me.

AliceMurray:
Hey,

I implemented this feature based on your ideas. See pull request #86 for details.

RagnarGrootKoerkamp:
I've just implemented some version of this for BAPCtools: https://github.com/RagnarGrootKoerkamp/BAPCtools/issues/35.

Happy to change that to whatever structure is decided upon eventually.

PR #86, opened 2018-07-26
Implement bad inputs for input/output format validators

Alicipy:
Now it's possible to specify invalid inputs and outputs which should be declined by input/output format validators. See #51 for details.

Usage:
 - Input Format Validator:  place .in files into a folder input_format_validators/bad_inputs, this input files should be declined by input validators
 - Output Format Validator: place {testcasename}.ans.wrong-{something} into data/{category}/ for wrong answers to inputfile {testcasename}.in. This wrong answers should be declined by the output validators. Interactive problems are currently not supported. A warning will occur if the inputfile doesn't exist (in case of misspelling)

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -1013,12 +1013,8 @@ class OutputValidators(ProblemAspect):
                     self.warning('Output validator check with wrong answers for interactive problems is currently not supported (skipping)')
                     return
 
-                if self._problem.config.get('validation') == 'default':
-                    self.warning('Output validator check with wrong answers and default validator is unnecessary (skipping)')
-                    return
-
             for wafile in wrong_answers:
-                for val in self._validators:
+                for val in self._actual_validators():
                     feedbackdir = tempfile.mkdtemp(prefix='feedback', dir=self._problem.tmpdir)
                     status, runtime = val.run(wafile,
                                               args=[testcase.infile, testcase.ansfile, feedbackdir] + flags,
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -836,7 +836,7 @@ class InputFormatValidators(ProblemAspect):
                     self.error('Input format validator %s crashed on input %s' % (val, invalid_infile))
 
                 # If an input validator declines the input file, everything is fine and we break.
-                if os.WEXITSTATUS(status) == 42:
+                if os.WEXITSTATUS(status) == 43:
                     break
             else:
                 # Will only be executed if loop wasn't ended by break.
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -323,6 +323,7 @@ class TestCaseGroup(ProblemAspect):
 
         infiles = glob.glob(os.path.join(self._datadir, '*.in'))
         ansfiles = glob.glob(os.path.join(self._datadir, '*.ans'))
+        wafiles = glob.glob(os.path.join(self._datadir, '*.ans.wrong-*'))
 
         if self._parent is None:
             seen_secret = False
@@ -358,12 +359,20 @@ class TestCaseGroup(ProblemAspect):
                 if len(files) > 1:
                     self.warning("Identical input files: '%s'" % str(files))
 
+        wafiles_with_input = []
+
         for f in infiles:
             if not f[:-3] + '.ans' in ansfiles:
                 self.error("No matching answer file for input '%s'" % f)
+            wafiles_with_input.extend(glob.glob(OutputValidators.WA_GLOB % (f[:-3] + '.ans')))
+
         for f in ansfiles:
             if not f[:-4] + '.in' in infiles:
                 self.error("No matching input file for answer '%s'" % f)
+
+        for f in wafiles:
+            if f not in wafiles_with_input:
+                self.error("No matching input file for wrong answer '%s'" % f)
 
         for subdata in self._items:
             if subdata.matches_filter(args.data_filter):
@@ -930,6 +939,8 @@ class Graders(ProblemAspect):
 class OutputValidators(ProblemAspect):
     _default_validator = run.get_tool('default_validator')
 
+    WA_GLOB = "%s.wrong-*"  # %s is the answerfile to a test case, i.e. 1.ans
+
 
     def __init__(self, problem):
         self._problem = problem
@@ -983,8 +994,47 @@ class OutputValidators(ProblemAspect):
                     self.warning('%s gets AC' % (desc))
             os.unlink(file_name)
 
+        self._verify_invalid_outputs()
+
         return self._check_res
 
+    def _verify_invalid_outputs(self):
+        """Check that output validators decline invalid answer files"""
+
+        val_timelim = self._problem.config.get('limits')['validation_time']
+        val_memlim = self._problem.config.get('limits')['validation_memory']
+        flags = self._problem.config.get('validator_flags').split()
+
+        for testcase in self._problem.testdata.get_all_testcases():
+            if not os.path.exists(testcase.infile):
+                # .in-file doesn't exist, already reported by testcase-check, so 'ignore' here
+                continue
+
+            wrong_answers = glob.glob(OutputValidators.WA_GLOB % testcase.ansfile)
+            if wrong_answers:
+                if self._problem.is_interactive:
+                    self.warning('Output validator check with wrong answers for interactive problems is currently not supported (skipping)')
+                    return
+
+                if self._problem.config.get('validation') == 'default':
+                    self.warning('Output validator check with wrong answers and default validator is unnecessary (skipping)')
+                    return
+
+            for wafile in wrong_answers:
+                for val in self._validators:
+                    feedbackdir = tempfile.mkdtemp(prefix='feedback', dir=self._problem.tmpdir)
+                    status, runtime = val.run(wafile,
+                                              args=[testcase.infile, testcase.ansfile, feedbackdir] + flags,
+                                              timelim=val_timelim, memlim=val_memlim)
+                    res = self._parse_validator_results(val, status, feedbackdir, testcase)
+                    shutil.rmtree(feedbackdir)
+                    if res.verdict != 'AC':
+                        # One output validator declined this wrong answer, so stop validating
+                        break
+                else:
+                    # Will only be executed if loop wasn't ended by break
+                    # No output validator declined wrong answer, this is an error.
+                    self.error("Wrong answer file %s was not declined by output validators" % wafile)
 
     def _parse_validator_results(self, val, status, feedbackdir, testcase):
         custom_score = self._problem.config.get('grading')['custom_scoring']
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -359,19 +359,16 @@ class TestCaseGroup(ProblemAspect):
                 if len(files) > 1:
                     self.warning("Identical input files: '%s'" % str(files))
 
-        wafiles_with_input = []
-
         for f in infiles:
             if not f[:-3] + '.ans' in ansfiles:
                 self.error("No matching answer file for input '%s'" % f)
-            wafiles_with_input.extend(glob.glob(OutputValidators.WA_GLOB % (f[:-3] + '.ans')))
 
         for f in ansfiles:
             if not f[:-4] + '.in' in infiles:
                 self.error("No matching input file for answer '%s'" % f)
 
         for f in wafiles:
-            if f not in wafiles_with_input:
+            if f.split('.ans.wrong-')[0] + '.in' not in infiles:
                 self.error("No matching input file for wrong answer '%s'" % f)
 
         for subdata in self._items:
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -795,8 +795,44 @@ class InputFormatValidators(ProblemAspect):
 
             os.unlink(file_name)
 
+        self._verify_invalid_inputs()
+
         return self._check_res
 
+    def _verify_invalid_inputs(self):
+        """Check that input format validators decline invalid input files
+           given in input_format_validators/bad_inputs
+        """
+
+        path_to_invalid_inputs = os.path.join(self._problem.probdir, 'input_format_validators', 'bad_inputs')
+
+        # verify only if invalid inputs are given, otherwise nothing to check in this function
+        if not os.path.exists(path_to_invalid_inputs):
+            return
+
+        # verify that invalid inputs are given in a directory, not a file
+        if not os.path.isdir(path_to_invalid_inputs):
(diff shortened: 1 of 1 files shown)


problemtools/verifyproblem.py
```diff
@@ -947,8 +994,47 @@ def check(self, args):
...
+                continue
+
+            wrong_answers = glob.glob(OutputValidators.WA_GLOB % testcase.ansfile)
+            if wrong_answers:
+                if self._problem.is_interactive:
+                    self.warning('Output validator check with wrong answers for interactive problems is currently not supported (skipping)')
+                    return
+
+                if self._problem.config.get('validation') == 'default':
+                    self.warning('Output validator check with wrong answers and default validator is unnecessary (skipping)')
```
simonlindholm: FWIW, I could see use cases for this, like declaratively saying that a trivial answer isn't the answer to any test case without having to manually check all the .ans files (which might even be hard to do with `diff` with floats involved). So if it's easy to remove this check it would be nice to do so.

problemtools/verifyproblem.py
```diff
@@ -947,8 +994,47 @@ def check(self, args):
...
+        val_timelim = self._problem.config.get('limits')['validation_time']
+        val_memlim = self._problem.config.get('limits')['validation_memory']
+        flags = self._problem.config.get('validator_flags').split()
+
+        for testcase in self._problem.testdata.get_all_testcases():
+            if not os.path.exists(testcase.infile):
+                # .in-file doesn't exist, already reported by testcase-check, so 'ignore' here
+                continue
+
+            wrong_answers = glob.glob(OutputValidators.WA_GLOB % testcase.ansfile)
```
simonlindholm: (I know this isn't in the hot path, but this linear-time, syscall-based, noop-in-99%-of-cases globbing still tears at me. Maybe we could loop over all wrong answers and pick out their associated testcases, or store wrong answer lists as part of testcase objects.)

problemtools/verifyproblem.py
```diff
@@ -947,8 +994,47 @@ def check(self, args):
...
+                                              timelim=val_timelim, memlim=val_memlim)
+                    res = self._parse_validator_results(val, status, feedbackdir, testcase)
+                    shutil.rmtree(feedbackdir)
+                    if res.verdict != 'AC':
+                        # One output validator declined this wrong answer, so stop validating
+                        break
+                else:
+                    # Will only be executed if loop wasn't ended by break
+                    # No output validator declined wrong answer, this is an error.
+                    self.error("Wrong answer file %s was not declined by output validators" % wafile)
```
simonlindholm: Like with the input validator, I don't like this duplication much (pretty complex code, too).

simonlindholm (commented):
Nice PR! This seems like a neat feature to have. @austrin maintains this repo, and is often quite late in responding, but in the meantime I made some comments in the hope that they're useful.

problemtools/verifyproblem.py
```diff
@@ -358,13 +359,21 @@ def check(self, args):
...
             if not f[:-3] + '.ans' in ansfiles:
                 self.error("No matching answer file for input '%s'" % f)
+            wafiles_with_input.extend(glob.glob(OutputValidators.WA_GLOB % (f[:-3] + '.ans')))
+
         for f in ansfiles:
             if not f[:-4] + '.in' in infiles:
                 self.error("No matching input file for answer '%s'" % f)
 
+        for f in wafiles:
+            if f not in wafiles_with_input:
```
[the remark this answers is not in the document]
AliceMurray: You are right, this is also much easier. I will change this (soon)

problemtools/verifyproblem.py
```diff
@@ -795,8 +804,44 @@ def modified_input_validates(applicable, modifier):
...
+            return
+
+        for invalid_input_filename in os.listdir(path_to_invalid_inputs):
+            invalid_infile = os.path.join(path_to_invalid_inputs, invalid_input_filename)
+            if not invalid_infile.endswith('.in'):
+                self.warning('Input file %s is not an input file' % invalid_input_filename)
+                continue
+
+            for val in self._validators:
+                status, _ = val.run(invalid_infile, args=None)
```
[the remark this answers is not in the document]
AliceMurray: I noticed that at implementing, I didn't use this feature myself, but I didn't know that this is important for IOI. Of course, implementing it at wrong input files like the output validator is possible. Like you said, it doesn't completely belong there. Another way is to have the same directory-structure like in /data, but I don't know if this is worthwhile.

I think a third opinion here would be nice.

problemtools/verifyproblem.py
AliceMurray: The problem here is that the default validator is not at self._validators (At least not in my tests), so every wrong output file is accepted.

Do you think that it's worthwhile to think about a solution here? With default validating, the trivial solution shouldn't be accepted because there is only one single solution normally.

problemtools/verifyproblem.py
AliceMurray: Like in the check if there is a corresponding input file? Yeah, nice idea. I thought this works fine and with 100 testcases at most, this won't be a big problem. Of course, your solution is better here.

problemtools/verifyproblem.py
AliceMurray: I think a function here would be nice. Maybe an abstract base class 'Validators' would help here, this would also make the check on noncompiling validators easier.

problemtools/verifyproblem.py
simonlindholm: It's not super important, but if you can get it to work at the same time as factoring output validation out into a single function it might be worth doing. Otherwise skip it.

(The trivial solution might be accepted if it turns out that it is in fact correct, because test data was generated in a broken way, which IME happens surprisingly often. Of course, this is why one writes greedy/stupid solutions and checks that they get WA, but it might be theoretically useful to declare that on a per-test case level. So yeah, not a super strong use-case, but not completely pointless either.)

problemtools/verifyproblem.py
simonlindholm: To abstract both input and output validators? I'm not sure, they don't share that much logic, and inheritance tends to be confusing. Maybe a helper function could be better (if anything; some small amount of duplication is fine IMO).

problemtools/verifyproblem.py
simonlindholm: Hm, `_actual_validators`?

problemtools/verifyproblem.py
```diff
@@ -795,8 +804,44 @@ def modified_input_validates(applicable, modifier):
...
+                if not os.WIFEXITED(status):
+                    self.error('Input format validator %s crashed on input %s' % (val, invalid_infile))
+
+                # If an input validator declines the input file, everything is fine and we break.
+                if os.WEXITSTATUS(status) == 42:
+                    break
```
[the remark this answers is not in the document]
simonlindholm: (fwiw, github hid this as "outdated" but it's still relevant)

problemtools/verifyproblem.py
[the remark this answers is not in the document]
simonlindholm: I don't know what the best choice is here either. Agreed that a third opinion would be nice.

problemtools/verifyproblem.py
AliceMurray: How could i miss that? Yes, that solves my problem. Works for default validation now.

Thank you very much!

problemtools/verifyproblem.py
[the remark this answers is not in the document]
meisterT: Simon, what flags would you pass here?

IMHO, the data directory should be cluttered as less as possible. The bad inputs don't have any relationship to the other test cases so they don't belong in here.
The bad outputs are related to a single test case so it makes more sense to store them next to the real thing although storing them in the output_validator section with the same data/{sample,secret,whatever} structure may also be an option to clearly separate between test data relevant for the contest(ants) and only for testing the validators.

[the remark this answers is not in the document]
simonlindholm: Flags either has to be `None` or taken from a .yaml file. I don't know the best option either.

I think the solution I'd lean towards is to put everything in `data/`. Say we add a directory to `data/validator-tests` with the same structure as `secret` and `sample`, and additionally files `*.{in,ans}.{invalid,valid}` (or `-wrong` or `-wrong-*` or whatever). This gives us input- and output validator flags, makes the structure uniform for both sorts of tests, is somewhat self-explanatory when you see it for the first time, and also makes it possible to add output validator tests without adding them as actual test cases. (For the Swedish IOI qualifiers for instance, we keep the `secret/` directory entirely auto-generated, while a `validator-tests` directory would be under version control.) We might want to support `.ans.invalid` in the actual data directories as well, of course.

FWIW, I recently tried out Polygon and it has the following UI for validator tests (same for input and output):
image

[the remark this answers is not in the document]
AliceMurray: After some testing I think that putting everything in data with the same structure as the other testcases is kind of hard. We would have to exclude this directory from the other testcases, implement that data/testcase-validator/sample should use the same flags as sample, think if wrong-answer-files should be able to refer to input-files from the testfiles and more. That is all possible, but require some changes on the current testcase class (like expand it with wrong answers, but that couldn't model invalid inputs) or other major restructures.

Should we check if the current state is good enough for a merge, or check some time more if we can fulfill all requested features here?

austrin:
I've been thinking a bit about how to incorporate this.  I agree with @meisterT about not cluttering the data directory too much.  I also find the `input_format_validators/bad_inputs` a bit problematic, because subdirectories in `input_format_validators` typically means a program.

My suggestion would be to have a new top level directory called something like `validator_tests`.  There, one can place a whole test case group hierarchy similar to the one under `data/`, and in particular have:
- `testdata.yaml` files controlling the input/output validator flags used for each test case group
- `X.in`, `X.ans`, `X.out.wrong` triples of files to test the output validator.
- `X.in.wrong` files to test the input validator.

Some downsides of this would be:
- It doesn't allow for adding output validator tests to the actual test cases (without making copies of them).  I personally think this is fine.
- Even if one can test the validators with different flags, if one wants to test them with the same groups as the actual test case groups of the problem, one would have to make sure that this parallel test case group tree is identical to the real one, which sounds annoying.  On the other hand I think problems with complex test case group structures often has the data tree generated, and then that generator can equally well generate an identical tree in `validator_tests/`.  @simonlindholm is this hypothesis correct?

simonlindholm:
That sounds like a good suggestion to me.

> @simonlindholm is this hypothesis correct?

It's true for testdata_tools, yes. I don't know if there is anyone else out there doing test groups with the problem format. One could also symlink testdata.yaml files if doing things manually. Or .in files.

godmar:
I've been looking for this for a long time, particularly for input validators.  Personally, I would just place them under data/badinput/*.in

I wasn't aware that there are input validator flags. Are they documented somewhere?

godmar:
Ok. FWIW, does Per @austrin suggest to name the directory data/validator_tests ?  If so, a risk is that it's not clear that all of these must be rejected.  I'd rather have something explicit like 'badinput' or even 'must_reject' - similar to how 'submissions/{accepted|wrong_answer|...}' makes it clear what goes there.

austrin:
@godmar the (current, still not completely finalized) plan is a directory `/validator_tests/` under the root of the problem package, with file names clearly indicating whether the expected result is that the input/output validator accepts/rejects, along the lines of what is described in https://github.com/Kattis/problemtools/pull/86#issuecomment-422722383

meisterT:
What's necessary to move the finalize the plan? I think your proposal is good and the downsides are not too bad.

AliceMurray:
The described format of @austrin sound pretty good. I have lost sight of this feature, I hope I have some time after NWERC and can implement this. So, technically, the finalization of this plan is just time :D

godmar:
I think it's good enough and it's definitely better than the status quo, although I aesthetically dislike 

- double suffixes like .in.wrong
- mixing input and output validators in the same directory because they have, IMO, little/nothing in common
- not staying under /data since we're talking about input data to the problem, and a separate root directory is to easy to overlook especially for new problem setters copying from a skeleton.

meisterT:
Did you talk about that this weekend in person? I totally forgot to bring it up but would love to use this feature.

austrin:
Alas no, I also forgot.  @AliceMurray if you feel like working on this in the near future let me know, otherwise I'm probably going to go ahead and do this now that there is a reasonable concrete plan.

AliceMurray:
Well, I also forgot discussing this in person this weekend.

If you want to, you can implement this, I won't have time in the near future. If you need someone to test it, let me know.

AliceMurray:
@austrin do you have already implemented something or need some help? In about a month, we have a small onside contest here and could test this feature on our problemset for you (under real conditions)

austrin:
@AliceMurray did not get around to it.  Would you have time to also implement it, or only to test it?  If the latter I can make an effort to get it implemented.

AliceMurray:
@austrin Currently, preparing the contest, studying and work is taking a lot of time. I think testing will be done by using it in our contest preperation, At the moment, I won't find time to implement it myself. I think it would be better if you know someone or can make an effort to implement it.

RagnarGrootKoerkamp:
I'm also thinking about how to do this for [BAPCtools](https://github.com/RagnarGrootKoerkamp/BAPCtools). So the current proposal is (just summarizing @austrin above)

Bad input:
`validator_tests[/test_groups]*.in.wrong`

Bad output:
`validator_tests[/test_groups]*.in`
`validator_tests[/test_groups]*.ans`
`validator_tests[/test_groups]*.out.wrong`

Regarding not being able to directly test on the actual testdata: I don't typically do this anyway. Usually I just make a few small hand-written inputs/output with bugs in them.

One thing I _slightly_ dislike is the mixing of bad input and output. Personally I would prefer to have something like a `validator_tests/bad_input/*.in` and `validator_tests/bad_ouput/*.{in,ans,out}`, but the current proposal is also good. Also, a drawback of separate directories for bad input and bad output would be that the test case groups have to be duplicated again.

pehrsoderman:
@AliceMurray @austrin What is the status here? What is needed to get this PR moving forward?

RagnarGrootKoerkamp:
I have a version of this working in BAPCtools:
- All bad IO is in `data/bad/`.
- Bad inputs go in `data/bad/*in`, and _do not_ have a corresponding `.ans`.
- Pairs of `data/bad/X.{in,ans}` indicate that `X.ans` is a bad output for `X.in`.

I don't really like the not-so-obvious mixing of bad inputs and outputs here though, so separating into something like `data/bad_input` and `data/bad_output` is probably nicer.

One benefit of keeping things in the `data/` directory is that it will work better in combination with `generators/generators.yaml`. In particular, I now have support for writing the contents of (per testgroup) `testdata.yml` in the `generators.yaml`. This means that it's possible to replicate a directory structure without doing too much duplication of code/config, since yaml objects can be reused.

I don't see too much value in allowing e.g. `.wrong.ans` right next to a `.in` in the `data/{sample,secret}/` tree. Usually I only add a few bad outputs for one/some small minimal testcases, where creating a new `.in` or possibly copying one to `data/bad` is fine.

PR #87, opened 2018-07-26
Fix input format validator skip by non-compiling validators

AliceMurray:
If a validator doesn't compile, the next one won't be compiled because the iteration skips one step. Copying the list first fixes this problem.

Additionally, the output validator is modified to have the same behaviour


austrin:
:+1: good catch!

PR #88, opened 2018-08-21
Remove -i flag from html tidy

### LINKED ISSUE #89: don't use -i with html tidy
HTML tidy with the `-i` option can introduce unwanted newlines and spaces inside of `<pre>` blocks.  This can cause problem2html to produce sample data that looks incorrect (has leading spaces, e.g.).

Here's a minimal example that gets messed up by `tidy -i`:

```
<!DOCTYPE html>
<html><head><title>test</title></head><body>
<pre>aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa</pre>
</body>
</html>
```
Running `tidy -i` on this input produces:
```
<!DOCTYPE html>
<html>
<head>
  <meta name="generator" content=
  "HTML Tidy for HTML5 for Linux version 5.2.0">
  <title>test</title>
</head>
<body>
  <pre>
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa</pre>
</body>
</html>
```
If you remove one of the "a" characters, then the problem disappears (the `<pre>` remains on one line).

The problem seems to occur when the pre input is too long.

I read a bit online -- it looks like this behavior is a known "feature" of HTML tidy. See this example which does exactly what we do not want -- adding a newline to the beginning of a `<pre>` section:
http://www.html-tidy.org/documentation/#featured-options-and-solutions
So... I think the "best" fix is just to remove `-i` from when we invoke tidy.

ghamerly:
Fixes #89 

Using this can introduce unwanted linebreaks and spaces inside `<pre>`
blocks. See the first example under
http://www.html-tidy.org/documentation/#featured-options-and-solutions

### DIFF
--- a/problemtools/problem2html.py
+++ b/problemtools/problem2html.py
@@ -74,7 +74,7 @@ def convert(problem, options=None):
             os.remove('.paux')
 
         if options.tidy:
-            os.system('tidy -utf8 -i -q -m %s 2> /dev/null' % destfile)
+            os.system('tidy -utf8 -q -m %s 2> /dev/null' % destfile)
 
         if options.bodyonly:
             content = open(destfile).read()


austrin:
Closing this as wontfix since newer tidy solves it.

PR #91, opened 2018-09-19
Some more validation for scoring problems

simonlindholm:
Also split out from #81.

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -182,7 +182,7 @@ class TestCase(ProblemAspect):
             return False
         return True
 
-    def run_submission(self, sub, args, timelim_low=1000, timelim_high=1000):
+    def run_submission(self, sub, args, timelim_low, timelim_high):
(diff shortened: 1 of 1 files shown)


problemtools/verifyproblem.py
```diff
@@ -379,23 +389,23 @@ def check(self, args):
         return self._check_res
 
 
-    def run_submission(self, sub, args, timelim_low, timelim_high):
+    def run_submission(self, sub, args, timelim_low, timelim_high, subinfo):
```
austrin: I am not happy about tacking on this subinfo argument down the callstack just to be able to prettyprint the error.

Remove this, and simply print `sub` for now (we lose the "ACR submission" part but that's OK for now).

simonlindholm: The subinfo argument is also used by #81 for more essential purposes, if that makes a difference. But I can remove it if we want to merge this early.

simonlindholm: I don't know how to parse your silence, but removed anyway.

PR #94, opened 2018-10-09
update to VIVA 2.1

ghamerly:
I pulled these directly from http://viva.vanb.org/; see that page for the change log to VIVA 2.1.

This release runs faster (I believe vanb said in particular regex was faster?), and properly rejects numbers with (extra) leading zeros.

I have tested this on a few, but not many, problems that have .viva validators.


niemela:
@ghamerly Is there any reason not to merge this?

If not, and you rebase, I can merge. Also, you could merge...

PR #98, opened 2018-10-24
Add dependencies on Fedora to README.md

cipherboy:
I was able to get all three programs working on Fedora 28 with these sets of dependencies. They're not necessarily complete as I started by translating the Ubuntu dependencies (and as this is a development machine, I had several dependencies already installed). With the exception of all the texlive packages missing from the Ubuntu dependencies, everything seemed to work well. Also, Fedora doesn't package `plastex` so I recommended installing that via `pip install --user`.

As a heads up, I'm not sure the future plans of this project, but you're looking at about a two year window for Fedora [dropping full support for Python 2](https://fedoraproject.org/wiki/FinalizingFedoraSwitchtoPython3). Fedora 29 will hopefully be released next week, Fedora 30 will drop support for python2 packages, and 33 looking like it will drop python2 completely. (Roughly 6 month release cycle). Fedora 30 is of interest if you want to see `problemtools` packaged in Fedora, otherwise for anyone running `problemtools` on Fedora, the Fedora 33 release is of interest.

I haven't looked into how much work porting problemtools to python3 would be, just figured I'd mention it as a heads up. I'd assume other distributions have similar plans. :)

`Signed-off-by: Alexander Scheel <alexander.m.scheel@gmail.com>`

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,14 @@ and a LaTeX installation.  In Ubuntu, the precise dependencies are as follows:
 
     libboost-regex1.54.0, libc6 (>= 2.14), libgcc1 (>= 1:4.1.1), automake, libgmp-dev, libgmp10, libgmpxx4ldbl, libstdc++6 (>= 4.4.0), python (>= 2.7), python (<< 2.8), python:any (>= 2.7.1-0ubuntu2), python-yaml, python-plastex, texlive-latex-recommended, texlive-fonts-recommended, texlive-latex-extra, texlive-lang-cyrillic, tidy, ghostscript
 
+On Fedora, these dependencies can be installed with:
+
+    sudo dnf install boost-regex gcc gmp-devel gmp-c++ python2 python2-pyyaml texlive-latex texlive-collection-fontsrecommended texlive-fancyhdr texlive-subfigure texlive-wrapfig texlive-import texlive-ulem texlive-xifthen texlive-overpic texlive-pbox tidy ghostscript
+
+Followed by:
+
+    pip2 install --user plastex
+
 The problem tools have not been tested on other platforms.  If you do
 test on another platform, we'd be happy to hear what worked and what
 did not work, so that we can write proper instructions (and try to


cipherboy:
(Should I target `develop` or `master` with this change? Currently I'm targeting `master` but it appears that `develop` is default on GitHub so I'm happy to rebase and switch).

simonlindholm:
`develop`, please -- `master` is only for releases. Otherwise the changes look good to me!

I think porting to Python 3 is probably not that much work, though there are other tools that depend on problemtools (e.g. in Kattis) which makes things a bit more complex. We'd want to make it compatible with both Python 2 and 3 as a first step, I imagine.

cipherboy:
@simonlindholm Yes, targeting both would be good. I haven't looked at what's required, only a user of `problemtools` for now. :)

This is targeting develop now. You might want to re-run Travis as it thinks it is against `master` still. Thanks!

simonlindholm:
> You might want to re-run Travis as it thinks it is against master still.

I don't know how to do that... on the other hand, I don't think a README update could possibly cause a failure. Thanks for the PR!

cipherboy:
On Wed, Oct 24, 2018, 08:47 Simon Lindholm <notifications@github.com> wrote:

> You might want to re-run Travis as it thinks it is against master still.
>
> I don't know how to do that... on the other hand, I don't think a README
> update could possibly cause a failure. Thanks for the PR!
>

Welcome! Glad that it worked on Fedora without many modifications. :)

For future reference, you can click on the Travis details in the GitHub PR,
which will take you to the overview of the Travis job(s) that were run. On
that page there's a button which says restart job. Usually you have to be a
project administrator/have commit rights for Travis to let you restart
them. I agree that the README shouldn't affect build results but I was
curious if Travis would pick up that it's against the develop branch
instead of master if it was just restarted instead of a push update. :)

It looks like you're only using a single task here so the button will
likely be in the top block on the right side. If you're using a more
complex build matrix, you can choose which jobs to restart or restart them
all.

simonlindholm:
Ah, yes, thanks. (I looked there before, but I had missed to log in.) I restarted it, and the UI still claims it's against master, though the commands executed are against the PR commit, not the merged state, so it shouldn't matter.

(https://travis-ci.org/Kattis/problemtools/builds/445636970. It also says it fails to `git fetch origin +refs/pull/98/merge` (`fatal: Couldn't find remote ref refs/pull/98/merge`), but that's probably expected since I merged it.)

PR #100, opened 2018-10-31
Allow Input and Output latex environments

RagnarGrootKoerkamp:
This allows for wrapping the Input and Output sections in an environment. Instead of hardcoding the `\section*{Input}` in the latex source, you can now also write:
```
\begin{Input}
The input is ...
\end{Input}
```

We use this in our BAPC repository because it gives some more flexibility. Here, I just map the begin of the environment to create a new uncounted section, like the default behaviour.


austrin:
Seems OK to me, but can you drop the weird merge commit from the PR?

RagnarGrootKoerkamp:
Thanks! Removed the merge commit from the PR.

PR #101, opened 2018-10-31
Let verifyproblem.py accept 1 or more problemdirs

ludopulles:
Allow verifyproblem.py to work on multiple problem directories.
This is useful to run it on, for example, a whole problemset.

It still gives an error when no problemdirs are supplied.


austrin:
Looks fine to me.

PR #106, opened 2018-11-23
remove dead code checking for pdf_ok

ghamerly:
This code does not appear to be doing anything useful. Its meaning is lost to time (pre-conversion to git).


austrin:
It used to be the case that if pdf conversion failed then the check for html conversion wouldn't even run (because odds are that will fail for the same reason anyway), i.e., `pdf_ok` would be set to false in the two error cases.  But I'm fine with removing it, because there are actually a few cases where pdf conversion will fail and html conversion will work.  Feel free to merge unless you think the description of the original purpose somehow warrants resurrecting that behaviour instead.

PR #110, opened 2018-12-09
fix bash run scripts

vaclavblazej:
I noticed that using problemtools via bash scripts did not support checking more than one problem definition. So when running `verifyproblem dog/ report/` the result is `verifyproblem.py: error: unrecognized arguments: report/`

This issue is fixed by the arguments from passing the arguments to the program with quotes: `"$@"`

---

Second issue was that the invoked python code changed depending on where the user was located.

This could be explained by a strange order of expanding the bash script location. The path is now first canonicalized and then we go up two levels to get to the proper folder. Note that `realpath` should be used instead of `readlink -f` but I am not sure about cross-platform compatibility, so this is not changed.

Hope this helps :)

### DIFF
--- a/bin/problem2html.sh
+++ b/bin/problem2html.sh
@@ -5,5 +5,6 @@
 # installing problemtools on the system properly, this script should
 # not be used.
 
-export PYTHONPATH=$(readlink -f $(dirname $0)/..):$PYTHONPATH
-exec python2 -m problemtools.problem2html $@
+export PYTHONPATH
+PYTHONPATH="$(dirname "$(dirname "$(readlink -f "$0")")"):$PYTHONPATH"
+exec python2 -m problemtools.problem2html "$@"
--- a/bin/problem2pdf.sh
+++ b/bin/problem2pdf.sh
@@ -5,5 +5,6 @@
 # installing problemtools on the system properly, this script should
 # not be used.
 
-export PYTHONPATH=$(readlink -f $(dirname $0)/..):$PYTHONPATH
-exec python2 -m problemtools.problem2pdf $@
+export PYTHONPATH
+PYTHONPATH="$(dirname "$(dirname "$(readlink -f "$0")")"):$PYTHONPATH"
+exec python2 -m problemtools.problem2pdf "$@"
--- a/bin/verifyproblem.sh
+++ b/bin/verifyproblem.sh
@@ -5,5 +5,6 @@
 # installing problemtools on the system properly, this script should
 # not be used.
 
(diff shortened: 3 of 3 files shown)


vaclavblazej:
Well, that just broke everything ... will look into that and reopen this when it works again.

vaclavblazej:
Gone into the rabbit hole of errors and found out that the fix was not to blame. I forgot to run `make` on the clean copy of the problemtools.

austrin:
Thanks!  Two questions:

1. For the first one, this should only make a difference if you have spaces in your arguments.  So I'm confused by your example of `verifyproblem dog/ report/`, that should already be working?  But adding the quotes is definitely the right thing.

2. For the second one, can you give an example of a setup where it currently fails?  I can see how it fails with paths involving spaces again due to missing quotes, but that's a different thing (which should be fixed).  I do find the double-dirname a bit annoying to read.

vaclavblazej:
Two answers:

1. This puzzled me as well, but it became fixed, so I assumed this was a 'bad' testcase. In reality this was caused by the second issue and the fact, that my global installation was installed a while ago, and this feature was not present in that version yet.

2. If we install version using pip and then clone this repository and create symbolic link to the bash then we get different `problemtools.verifyproblem` module started depending on if the `PYTHONPATH` was correctly set.

Since the original version first went up 2 levels and then used realpath it gradually got something like this:
```
/home/blazeva1/bin/verifyproblem2 // original
/home/blazeva1/bin   //1 dirname
/home/blazeva1/bin/.. // added /..
/home/blazeva1 // realpath
```
However the symbolic link should be canonified first to get the real location of the script and then go up 2 levels (note the order matters here).
```
/home/blazeva1/bin/verifyproblem2 //original
/home/blazeva1/bin/problemtools/bin/verifyproblem.sh //realpath
/home/blazeva1/bin/problemtools/bin //first dirname
/home/blazeva1/bin/problemtools // second dirname
```
This caused the pip installation to be found first and so the script ran the wrong version.

vaclavblazej:
I guess the realpath was originally used to evaluate the `/..` part of the path. If you want to avoid "annoying" dirnames then `PYTHONPATH="$(realpath "$(dirname "$(realpath "$0")")/.."):$PYTHONPATH"` could be used. Inner realpath to evaluate symbolic links and outer realpath to evaluate `/..` part of the path. But this does not seem to help that much ...

austrin:
Ah, I see, it happens when you make a symlink of just the script (rather than the whole `bin` directory, which was what I was thinking of).  I agree that this is a bug to be fixed, and your fixes seem good, so will merge.

PR #120, opened 2019-03-15
Fix SIGPIPE-related race for interactive problems

simonlindholm:
See the updated comment.

### DIFF
--- a/support/interactive/Makefile
+++ b/support/interactive/Makefile
@@ -5,4 +5,4 @@ clean:
 	rm -f $(PROGRAM)
 
 %: %.cc
-	g++ -O3 -o $@ $<
+	g++ -O3 -Wall -o $@ $<
--- a/support/interactive/interactive.cc
+++ b/support/interactive/interactive.cc
@@ -13,6 +13,7 @@
 
 #define NOFD -1
 
+const int EXITCODE_AC = 42;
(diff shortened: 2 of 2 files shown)


ghamerly:
> One question I have, though, is whether it is necessary to `close(fromuser[0])` at all? If we do not, then the submission would not die on `SIGPIPE`, right?

I retract the question... here's more of my thinking about it.

When the validator exits first with a non-AC judgement, the submission can then finish in one of three ways: cleanly (within time without crashing), on RTE (e.g. caused by `SIGPIPE`), or killed when it gets TLE.

So if we `close(fromuser[0])`, we make it more likely to get the second (RTE) due to the possibility of `SIGPIPE`. But since we would take the validator's judgement anyway (since the validator exited first and with non-AC), how the submission exited is not really an issue.

So, my understanding is fine to `close(fromuser[0])` in this situation, but also I think strictly speaking it's unnecessary. However, it can possibly cut down on some TLE situations (saving judging time?).

niemela:
> However, it can possibly cut down on some TLE situations (saving judging time?).

Right, and that is good.

simonlindholm:
Exactly. This is similar to how we close write ends when validator/submission exits.

austrin:
LGTM and I've tested it.  Could probably be more extensively tested but I think it is safe enough to merge so will do so.

PR #121, opened 2019-04-03
Update default grader with first_error, worst_error, ignore_sample, prefer_accepted

simonlindholm:
Recently added to the spec after some discussion with @niemela. first_error is rarely useful, but can be good for avoiding timing leaks and for analytics in judges that don't do lower-priority judging of unnecessary testcases. worst_error is a better name for no_errors. ignore_sample is useful for scoring problems (unless secret/ has a grader that always outputs Accepted (0), as we're doing with PO; a Rejected status would help with that).

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -956,7 +956,7 @@ class Graders(ProblemAspect):
             graders = self._graders
 
         grader_input = ''.join(['%s %s\n' % (r.verdict, 0 if r.score is None else r.score) for r in sub_results])
-        grader_output_re = r'^((AC)|(WA)|(TLE)|(RTE))\s+[0-9.]+\s*$'
+        grader_output_re = r'^((AC)|(WA)|(TLE)|(RTE)|(JE))\s+[0-9.]+\s*$'
         verdict = 'AC'
         score = 0
 
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -385,6 +385,12 @@ class TestCaseGroup(ProblemAspect):
         if self.config['grading'] == 'default' and Graders._default_grader is None:
             self._problem.graders.error('%s has default grading but I could not find default grader' % self)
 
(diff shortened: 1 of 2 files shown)


simonlindholm:
Now updated with prefer_accepted and some validation in verifyproblem.

austrin:
Is it still WIP or is it ready to merge?

What exactly is the use case of `prefer_accepted`?

simonlindholm:
Ready to merge! prefer_accepted is for subgroups, where the secret/ directory should get Accepted if any subgroup does, but otherwise use something like first_error or worst_error.

austrin:
While flying I was thinking that `prefer_accepted` is maybe not an optimal name.  Not sure what would be, but wouldn't something like `accept_if_any_accepted` or `one_accepted_sufficient` be a bit better (more descriptive)?

ghamerly:
> While flying I was thinking that `prefer_accepted` is maybe not an optimal name. Not sure what would be, but wouldn't something like `accept_if_any_accepted` or `one_accepted_sufficient` be a bit better (more descriptive)?

`at_least_one_accepted`? `one_or_more_accepted`? `any_accepted`?

austrin:
@ghamerly I think your suggestions are not great -- the point of the flag is to say that "accept even if there is just a single AC sub-result".  So e.g. `at_least_one_accepted` sends the wrong message because you need at least one accepted to get AC both with and without the flag.

I think my preference is the very explicit `one_accepted_sufficient`.  But I guess `prefer_accepted` is fine.  @simonlindholm what do you think?

simonlindholm:
I think I'd be happy with `accept_if_any_accepted`.

niemela:
I'd accept (pun intended) `accept_if_any_accepted`, but I'd prefer (yes...) `prefer_accepted`.

simonlindholm:
Renamed to `accept_if_any_accepted` (so requires another spec update). Let's try to get this merged.

niemela:
Spec is updated. Go ahead and merge!

PR #124, opened 2019-05-03
Correctly capitalize Mainclass, only change first letter

mpsijm:
Python's string `capitalize` function indeed capitalizes the first character of the string, but it also converts all other characters in the string to lowercase. For example, "e_CorrecT" would be `capitalize`d to "E_correct".

However, `kotlinc` only capitalizes the first letter of the class and leaves the other characters unchanged. For example, "e_CorrecT" will be compiled by Kotlin to "E_CorrecTKt". The definition of the `Mainclass` variable has been updated to fix this (the "Kt" suffix is handled in config/languages.yaml).

### DIFF
--- a/problemtools/run/source.py
+++ b/problemtools/run/source.py
@@ -72,7 +72,7 @@ class SourceCode(Program):
             self.mainfile = self.src[0]
 
         self.mainclass = os.path.splitext(os.path.basename(self.mainfile))[0]
-        self.Mainclass = self.mainclass.capitalize()
+        self.Mainclass = self.mainclass[0].upper() + self.mainclass[1:]
 
         self.binary = os.path.join(self.path, 'run')
 


mpsijm:
Build error is not caused by me. From the build log of [build #213 (the last commit to the develop branch)](https://travis-ci.org/Kattis/problemtools/builds/519873843):

> Using the default Python version 2.7. Starting on 2019-04-16 the default will change to 3.6. If you wish to keep using this version beyond this date, please explicitly set the Python value in configuration.

simonlindholm:
`.upper()` maybe makes more sense than `.capitalize()`? Either way, looks fine to me, I can merge if you rebase since it looks like @austrin is being super slow.

mpsijm:
@simonlindholm Done :slightly_smiling_face: Thanks! :smile:

PR #127, opened 2019-05-06
Replace os.system with subprocess.call; add code_size() to Problem class

### LINKED ISSUE #125: BuildRun object has no attribute code_size
The code below fails with the error in the subject line when run on a problem with a a build/run submission.

https://github.com/Kattis/problemtools/commit/bf6d02bc2b1bb464b53b59eec0bb47117ae5c627#diff-8f07293bb3f09e73192d4a5816acb9dcR1250

### LINKED ISSUE #126: verifyproblem fails when ancestor directory has space in path name
If a problem is a part of a path with an ancestor directory with a space in the name (NB: not in the problem directory's name), veryifyproblem fails. This is because problem2pdf fails on a call to `os.system()`.

I'm working on a fix which replaces all `os.system` calls with `subprocess.call`.

ghamerly:
Two commits -- for two different issues

First commit fixes #126 
Second commit fixes #125 in a generic way

### DIFF
--- a/problemtools/ProblemPlasTeX/__init__.py
+++ b/problemtools/ProblemPlasTeX/__init__.py
@@ -1,6 +1,7 @@
 import re
 import os
 import shutil
+import subprocess
 import plasTeX.Renderers
 from plasTeX.Renderers.PageTemplate import Renderer
 from plasTeX.Filenames import Filenames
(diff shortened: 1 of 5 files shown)


simonlindholm:
The return values of `os.system` and `subprocess.call` are different -- os.system returns a C exit code which should be checked by `os.WIFEXITED` etc., while subprocess.call returns either a non-negative number to indicate a normal exit with that return code, or a negative value to indicate a signal.

ghamerly:
> The return values of `os.system` and `subprocess.call` are different -- os.system returns a C exit code which should be checked by `os.WIFEXITED` etc., while subprocess.call returns either a non-negative number to indicate a normal exit with that return code, or a negative value to indicate a signal.

I was worried about that, but could not find good documentation on it quickly (e.g. this link seems not to point out the differences clearly, even though it should: https://docs.python.org/2/library/subprocess.html#replacing-os-system). I'll make some updates.

austrin:
@ghamerly looks fine to me, can we merge?

ghamerly:
> @ghamerly looks fine to me, can we merge?

Yes, sorry for the slow response.

PR #132, opened 2019-08-23
Parenthesize if/else tuples to avoid a crash

### LINKED ISSUE #131: verifyproblem.py crashes on time limit warning
I get the following error when a TLE solution triggers a "sensitive to time limit" warning:

>   File "/u/evouga/icpc/problemtools/problemtools/verifyproblem.py", line 1272, in check_submission
    r1, r2 = result1, result2 if result1.verdict == result2.verdict else result1.verdict, result2.verdict
ValueError: too many values to unpack

I'm no Python expert but I believe there is a precedence problem with line 1272 of verifyproblem.py. I fixed the error locally with this change:

````
            r1, r2 = (result1, result2) if result1.verdict == result2.verdict else (result1.verdict, result2.verdict)
````

ghamerly:
Fixes #131

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -1269,7 +1269,7 @@ class Submissions(ProblemAspect):
             self.warning('%s got %s on sample: %s' % (desc, res.verdict, res))
 
         if result1.verdict != result2.verdict or result1.score != result2.score:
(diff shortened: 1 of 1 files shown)


simonlindholm:
I guess this may require another release, @austrin?

austrin:
@simonlindholm there's now a 1.20190901 release (which apart from this bugfix includes mostly complete Python 3 compatibility and interactive samples).

PR #134, opened 2019-09-27
Use -std=gnu++17

ghamerly:
This would make problemtools consistent with Kattis.

I can only think of one reason *not* to do this: ICPC rules still list `-std=gnu++14` (https://icpc.baylor.edu/xwiki/wiki/public/view/worldfinals/programming-environment)


niemela:
> I can only think of one reason _not_ to do this: ICPC rules still list `-std=gnu++14` (https://icpc.baylor.edu/xwiki/wiki/public/view/worldfinals/programming-environment)

I would expect/hope that that gets updated?

austrin:
The ICPC programming environment now has gnu++17.

PR #137, opened 2019-10-17
Require that graders exit with return code 0

simonlindholm:
Kattis checks this and gives judge errors on other status codes. 42 would be slightly more consistent, but it's not backwards compatible to require it.


austrin:
Looks good to me, I also updated the problem format specification to explicitly say that graders are expected to exit with 0, for good measure.

PR #139, opened 2019-10-18
Reset signal dispositions before running exec

simonlindholm:


simonlindholm:
I should probably add some short description to this... This came up when debugging interactive problem support before the ICPC (programs run through isolate already do the signal resets, so we had a fun Kattis/problemtools discrepancy). It's not very important, but it's still the right thing to do.

I tested that programs run with this code have their signal dispositions changed, and that verifyproblem still works on the examples.

PR #144, opened 2019-11-02
verifyproblem.py: check for invalid filename characters

### LINKED ISSUE #143: Flag invalid characters in data filenames
An issue came up at the MCPC 2019 where Kattis threw a judge error due to an '=' character in a judge input filename. It would be helpful is problemtools validated and complained about any invalid file names.

ghamerly:
Fixes #143, though we may want to debate what characters should be
whitelisted.

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -83,6 +83,7 @@ class ProblemAspect:
     warnings = 0
     bail_on_error = False
     _check_res = None
+    basename_regex = re.compile('^[a-zA-Z0-9][a-zA-Z0-9_.-]*[a-zA-Z0-9]$')
 
     @staticmethod
     def __append_additional_info(msg, additional_info):
@@ -123,6 +124,10 @@ class ProblemAspect:
     def debug(self, msg):
         logging.debug(': %s', msg)
 
+    def check_basename(self, path):
+        basename = os.path.basename(path)
+        if not self.basename_regex.match(basename):
+            self.error("Invalid name '%s' (should match '%s')" % (basename, self.basename_regex.pattern))
 
 class TestCase(ProblemAspect):
     def __init__(self, problem, base, testcasegroup):
@@ -154,6 +159,8 @@ class TestCase(ProblemAspect):
         if self._check_res is not None:
             return self._check_res
         self._check_res = True
+        self.check_basename(self.infile)
+        self.check_basename(self.ansfile)
         self.check_newlines(self.infile)
         self.check_newlines(self.ansfile)
         self._problem.input_format_validators.validate(self)
@@ -386,6 +393,8 @@ class TestCaseGroup(ProblemAspect):
         if self._check_res is not None:
             return self._check_res
         self._check_res = True
+
+        self.check_basename(self._datadir)
 
         if self.config['grading'] not in ['default', 'custom']:
             self.error("Invalid grading policy in testdata.yaml")
@@ -1374,7 +1383,6 @@ class Submissions(ProblemAspect):
             limits['time'] = timelim
 
         return self._check_res
-
 
 PROBLEM_PARTS = ['config', 'statement', 'validators', 'graders', 'data', 'submissions']
 


problemtools/verifyproblem.py
```diff
@@ -470,6 +470,16 @@ def check(self, args):
             if not f[:-4] + '.in' in infiles:
                 self.error("No matching input file for answer '%s'" % f)
 
+        # Check for filename characters which cause problems when judging; in
+        # particular, do not allow "=" in the filename as it causes problems
+        # in the way Kattis uses isolate. There are probably other characters
+        # we might allow, but we err on the side of safety.
+        test_filename_whitelist = set(string.letters + string.digits + '._-')
```
simonlindholm: `string.letters` -> `string.ascii_letters` for Python3 compatiblity

ghamerly:
> LGTM! Kattis's regex when uploading files looks like this, btw:

This check here is for test file names, not for source files, so I'm not sure what you're pointing out?

simonlindholm:
Oops, I skimmed through this too quickly. Anyhow, I think it's reassuring that it uses basically the same whitelist.

problemtools/verifyproblem.py
```diff
@@ -470,6 +470,16 @@ def check(self, args):
             if not f[:-4] + '.in' in infiles:
                 self.error("No matching input file for answer '%s'" % f)
 
+        # Check for filename characters which cause problems when judging; in
+        # particular, do not allow "=" in the filename as it causes problems
+        # in the way Kattis uses isolate. There are probably other characters
+        # we might allow, but we err on the side of safety.
```
austrin: I don't think comments like this belong here.  My view is that problemtools is primarily supposed to be an implementation of the problem format standard, and implementation details of Kattis like this one should not be relevant.

ghamerly: Thanks; when I was originally writing this I looked for a regex in the standard and did not find it, but @niemela pointed me to it. Fixed in acdcdb3.

problemtools/verifyproblem.py
```diff
@@ -470,6 +470,14 @@ def check(self, args):
             if not f[:-4] + '.in' in infiles:
                 self.error("No matching input file for answer '%s'" % f)
 
+        # Check that test data filenames are legal (according to the problem
+        # format standard).
+        filename_regex = re.compile('^[a-zA-Z0-9][a-zA-Z0-9_.-]*[a-zA-Z0-9]$')
```
austrin: There is something grating about this regex being recompiled for every test case group, even though it presumably has no noticeable performance impact.  Shouldn't this be a static/constant defined in a slightly less local scope?

ghamerly:
> Shouldn't we also check that test data group names match that regex?

Before I go too much further, let's discuss (perhaps not here?) what files we should be checking. The standard says "All file names for files included in the package must match the following regexp `[a-zA-Z0-9][a-zA-Z0-9_.-]*[a-zA-Z0-9]`" 
(https://clics.ecs.baylor.edu/index.php/Problem_format#General_Requirements)

I can easily add a check for all filenames in the problem package, but that's likely to be annoying. I.e. it will trigger on all files which have invalid characters (like `~` backup files, files that begin with a period (such as `.timelimit`) and such).

I've pushed such a solution (just as a starting point for discussion). It's simple. But like I said it's annoying.

austrin:
> Before I go too much further, let's discuss (perhaps not here?) what files we should be checking.

Right, sorry for the scope creep (I meant for this PR to stay within the test data but should have been more clear).

I agree with you that we should not check every single file appearing in the package.  I think it would make sense to only check files which are assigned meaning by the standard.   That's a somewhat complicated set though and we can add a few such checks to cover most things.  We don't need to make an effort to find all such places for this PR but can for now happily settle for the data files (and, if you want, the closely related test case groups).

niemela:
> I agree with you that we should not check every single file appearing in the package.

Right. This line in the spec can be read either as a requirement (i.e. all files inside the problem package root directory must follow this naming) or as a definition (i.e. files that do not follow the naming are not included in the package, meaning they can be ignored). We definitely want to allow users to store whatever that does not conflict with thing that are assigned meaning in the package.

ghamerly:
Okay, now we're checking all data files, according to a regex that is compiled once (in ProblemAspect) and can be used by any subclass. But it's currently only used on *.in/*.ans files.

austrin:
> Okay, now we're checking all data files, according to a regex that is compiled once (in ProblemAspect) and can be used by any subclass. But it's currently only used on .in/.ans files.

OK that's fine by me if you for some reason don't want to apply it to group names.

Can you clean up the commit history of the PR and then we can merge?

ghamerly:
> OK that's fine by me if you for some reason don't want to apply it to group names.

Can you clarify what you mean? Do you mean directory names? I'm fine with that if that's what you mean.

austrin:
Yes group names = directory names under `data/`

ghamerly:
I added test data groups, rebased, and squashed.

PR #153, opened 2019-11-26
Build debian for python3

pehrsoderman:
This PR makes us build a debian which contains modules for both python2 and python3. While fixing the test errors I did a quick cleanup of the type comparisons. Note that one check was specifically designed to look for a "long" type. This type does not exist in python 3. I removed this check, and added a check for string types instead in `test_languages.py`.  

This should pass CI once #147 is merged.
I think this could be consider a part of #116

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: python
-python: 2.7
+python:
+  - 2.7
+  - 3.7
 
 script: py.test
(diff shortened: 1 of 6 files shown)


austrin:
#147 is merged.  Can you rebase and update the PR?

pehrsoderman:
Done and passes CI now.

pehrsoderman:
Also, if I can get a release when this is merged, it would be helpful. ;)

austrin:
Will make a release later tonight.

austrin:
@pehrsoderman there is now a v1.20191126 release.

PR #157, opened 2019-12-04
Replace dependency on texlive-latex-recommended with texlive-plain-generic

### LINKED ISSUE #155: Problemtools depends on texlive-generic-recommended, but this package is about to be removed
Problemtools currently depends on `texlive-generic-recommended`. However, this package is about to be removed and replaced with `texlive-plain-generic`. There is a transitional dummy package in disco (19.04), but it gets dropped in eoan (19.10), which prevents problemtools from installing.

pehrsoderman:
texlive-latex-recommended is a transitional dummy package which is removed
in eoan (19.10). It only depends on texlive-plain-generic.

See https://packages.ubuntu.com/cosmic/texlive-generic-recommended

Fixes #155

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ RUN apt-get update && \
             texlive-fonts-recommended \
             texlive-lang-cyrillic \
             texlive-latex-extra \
-            texlive-latex-recommended \
+            texlive-plain-generic \
             tidy \
             vim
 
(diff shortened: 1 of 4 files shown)


austrin:
Can you update the documentation in the root README.md accordingly?

austrin:
As well as the `admin/docker/Dockerfile.minimal` file?

pehrsoderman:
@austin: I updated the docker file and README.md. I also updated the root docker file.

PR #161, opened 2019-12-19
Updated language config to match Kattis

pehrsoderman:
This updates the language config with two changes:

1. Use pypy instead of cPython to run python both for Python 2 and Python 3.
2. Include .jar files in the local directory for Java, to enable problems to have a .jar file.

### DIFF
--- a/problemtools/config/languages.yaml
+++ b/problemtools/config/languages.yaml
@@ -147,8 +147,8 @@ java:
     name: 'Java'
     priority: 800
     files: '*.java'
-    compile: '/usr/bin/javac -encoding UTF-8 -sourcepath {path} -d {path} {files}'
-    run: '/usr/bin/java -Dfile.encoding=UTF-8 -XX:+UseSerialGC -Xss64m -Xms{memlim}m -Xmx{memlim}m -cp {path} {mainclass}'
+    compile: '/usr/bin/javac -encoding UTF-8 -sourcepath {path} -d {path} -cp {path}/* {files}'
+    run: '/usr/bin/java -Dfile.encoding=UTF-8 -XX:+UseSerialGC -Xss64m -Xms{memlim}m -Xmx{memlim}m -cp {path}:{path}/* {mainclass}'
 
 javascript:
     name: 'JavaScript'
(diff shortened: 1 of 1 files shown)


problemtools/config/languages.yaml
```diff
@@ -147,8 +147,8 @@ java:
     name: 'Java'
     priority: 800
     files: '*.java'
-    compile: '/usr/bin/javac -encoding UTF-8 -sourcepath {path} -d {path} {files}'
-    run: '/usr/bin/java -Dfile.encoding=UTF-8 -XX:+UseSerialGC -Xss64m -Xms{memlim}m -Xmx{memlim}m -cp {path} {mainclass}'
+    compile: '/usr/bin/javac -encoding UTF-8 -sourcepath {path} -d {path} -cp {path}/* {files}'
```
meisterT: What is this change for?

austrin:
I do not view this *default* language config that ships with problemtools as something that is intended to be in sync with whatever open.kattis.com happens to be using at the moment, but rather as some reasonable defaults that lets a user get started.  Put differently, I don't think that *configuration* of a *specific* Kattis installation belongs in the problemtools code base -- such configuration (e.g. for open.kattis.com) should  be handled as documented in https://github.com/kattis/problemtools#configuration

Because the premise of "matching Kattis" is in my view flawed, I will close this.

(There is another discussion one can have about what the most reasonable defaults are, and there the choice of cpython vs pypy is not really clear-cut, but that's a different issue.)

niemela:
> Because the premise of "matching Kattis" is in my view flawed

Agreed, that's a flawed argument.

> I will close this.

That said, both of these changes makes sense on their own. I.e. not only because they are used on Kattis.

Is there any reason that CPython is a better default than Pypy? I would slightly lean towards the latter.

Allowing .jar files definitely seems like the correct thing to do to support "included code" properly. In fact, I would argue that it is simply an oversight that it has not already been done. That said, maybe there is some reason against?

problemtools/config/languages.yaml
pehrsoderman: It allows the problem to include a .jar, and dropping it in the problem directory during the compile, instead of just lose .java files. By using -cp java will look for them in the specified directory.

simonlindholm:
> I do not view this default language config that ships with problemtools as something that is intended to be in sync with whatever open.kattis.com happens to be using at the moment, but rather as some reasonable defaults that lets a user get started. Put differently, I don't think that configuration of a specific Kattis installation belongs in the problemtools code base -- such configuration (e.g. for open.kattis.com) should be handled as documented in https://github.com/kattis/problemtools#configuration

In that case, can we have some documentation e.g. in the problemtools README about what configuration to use to match Kattis? Or can there be a config flag one can set once and for all to keep up to date with the Kattis problemtools configuration? Preparing Kattis problems is the only thing I use problemtools for, and I'm sure that's true for many of us.

PR #167, opened 2020-02-23
Only AC submissions should have scores.

ghamerly:
Move the check for score after WA detection, so that WA does not get a score.

@simonlindholm do you have an opinion on this?


simonlindholm:
+1 from me. Could simplify `return SubmissionResult('WA', score=score, ...` to just `return SubmissionResult('WA', ...`.

ghamerly:
Thanks @simonlindholm; I took your suggestion, and also changed a few more things (including one return value which I think was a bug -- returning a `SubmissionResult` object from `grade()` when a pair is expected).

Issue #79, opened 2018-05-20
Clean install on Ubuntu 18.04 failing with make builddeb


FinnLidbetter:
I'm having trouble installing the problemtools. I am working from Ubuntu 18.04 and after installing only git and the listed build dependencies in debian/control I get the following error: `/usr/lib/python2.7/dist-packages/setuptools/dist.py:406: UserWarning: The version specified ('v1.20170420-15-ga52bdb2') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.`

Complete output on the command `make builddeb` is as follows.

finn@finn-VirtualBox:~/problemtools$ make builddeb
dpkg-buildpackage -us -uc -tc -b
dpkg-buildpackage: info: source package kattis-problemtools
dpkg-buildpackage: info: source version 1.20180426
dpkg-buildpackage: info: source distribution artful
dpkg-buildpackage: info: source changed by Per Austrin <austrin@kattis.com>
dpkg-buildpackage: info: host architecture amd64
 dpkg-source --before-build problemtools
 fakeroot debian/rules clean
make[1]: Entering directory '/home/finn/problemtools'
dh clean --with python2 --buildsystem=pybuild
   dh_auto_clean -O--buildsystem=pybuild
   dh_clean -O--buildsystem=pybuild
make[1]: Leaving directory '/home/finn/problemtools'
 debian/rules build
make[1]: Entering directory '/home/finn/problemtools'
dh build --with python2 --buildsystem=pybuild
   dh_update_autotools_config -O--buildsystem=pybuild
   dh_auto_configure -O--buildsystem=pybuild
I: pybuild base:217: python2.7 setup.py config 
/usr/lib/python2.7/dist-packages/setuptools/dist.py:406: UserWarning: The version specified ('v1.20170420-15-ga52bdb2') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  "details." % self.metadata.version
running config
   dh_auto_build -O--buildsystem=pybuild
I: pybuild base:217: /usr/bin/python setup.py build 
/usr/lib/python2.7/dist-packages/setuptools/dist.py:406: UserWarning: The version specified ('v1.20170420-15-ga52bdb2') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  "details." % self.metadata.version
running build
running build_support
[...]

PR #171, opened 2020-04-19
Init submodule from root Makefile

mikaelgoldmann:
Fixes: #79

Apparently relying on `dpkg-buildpackage` to run `support/Makefile` early
enough is not reliable. I haven't dug into why but rather added
init:ing the checkdata submodule to the root `Makefile`'s targets. Worst case, adding it for the `all` target is redundant given the `support/Makefile`

Tested as far as running `make builddeb` in clean clone from inside a `ubuntu:18.04` docker image (running on my macbook).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,10 @@
-all:
+all: checktestdata
 	make -C support
 
-builddeb:
+builddeb: checktestdata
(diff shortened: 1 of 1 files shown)


mikaelgoldmann:
This seems harmless and should fix build issues from a clean checkout. I will just merge it myself.

Issue #165, opened 2020-01-13
Docker image still uses Python 2


mpsijm:
First of all, thanks for the awesome work in #116, considering the EOL of Python 2 :smile:

However, the Docker image still seems to use Python 2, as shown below:
```
root@641e2610d3c4:/# cat /usr/bin/verifyproblem 
#! /usr/bin/python2
# EASY-INSTALL-ENTRY-SCRIPT: 'problemtools==1.20191126.post1','console_scripts','verifyproblem'
__requires__ = 'problemtools==1.20191126.post1'
import re
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(
        load_entry_point('problemtools==1.20191126.post1', 'console_scripts', 'verifyproblem')()
    )
```

Just changing the shebang from `/usr/bin/python2` to `3` did not work:

```
root@641e2610d3c4:/# python3 /usr/bin/verifyproblem 
Traceback (most recent call last):
  File "/usr/bin/verifyproblem", line 6, in <module>
    from pkg_resources import load_entry_point
ModuleNotFoundError: No module named 'pkg_resources'
```

P.S. various other scripts also still seem to use Python 2 in the shebang, like [setup.py](https://github.com/Kattis/problemtools/blob/develop/setup.py) and [problem2pdf.py](https://github.com/Kattis/problemtools/blob/develop/problemtools/problem2pdf.py).

mpsijm:
Closed by #174 :smile:

PR #174, opened 2020-06-11
Update ubuntu; move to python3; other updates

ghamerly:
This PR does a number of updates:
* problemtools scripts now depend only on python3 rather than python2 (e.g. verifyproblem, problem2html, problem2pdf) -- it removes support for running under python2 (NB: it does not remove python2 as a language for submissions)
* update docker images to run under Ubuntu 20.04 (from 18.04)
* uses pypy3 as the Python 3 interpreter in docker images
* replace libmozjs with nodejs in the language configuration

I have built docker images from this branch which are available here for testing:
* https://hub.docker.com/r/hamerly/problemtools-full
* https://hub.docker.com/r/hamerly/problemtools-icpc
* https://hub.docker.com/r/hamerly/problemtools-minimal
* https://hub.docker.com/r/hamerly/problemtools-build

I would appreciate testing and feedback if anyone has time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM ubuntu:20.04
 
 MAINTAINER austrin@kattis.com
 
@@ -14,10 +14,10 @@ RUN apt-get update && \
             libgmp10 \
             libgmpxx4ldbl \
             openjdk-8-jdk \
-            python-minimal \
-            python-pip \
-            python-plastex \
-            python-yaml \
+            python3-minimal \
+            python3-pip \
+            python3-plastex \
+            python3-yaml \
             sudo \
             texlive-fonts-recommended \
             texlive-lang-cyrillic \
@@ -26,4 +26,4 @@ RUN apt-get update && \
             tidy \
             vim
 
-RUN pip install git+https://github.com/kattis/problemtools
+RUN pip3 install git+https://github.com/ghamerly/problemtools@move-to-py3
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,4 +26,4 @@ RUN apt-get update && \
             tidy \
             vim
 
-RUN pip3 install git+https://github.com/ghamerly/problemtools@move-to-py3
+RUN pip3 install git+https://github.com/kattis/problemtools
--- a/admin/docker/Dockerfile.icpc
+++ b/admin/docker/Dockerfile.icpc
@@ -18,7 +18,7 @@ RUN apt update && \
         g++ \
         openjdk-11-jdk \
         pypy \
-        python3
+        pypy3
 
 # Install Kotlin
 WORKDIR /usr/local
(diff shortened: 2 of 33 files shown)


ghamerly:
I believe these changes are a superset of #168, so if we merge this, we should close that one.

problemtools/verifyproblem.py
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/env python2
+#! /usr/bin/env python3
 # -*- coding: utf-8 -*-
 from __future__ import print_function
```
thorehusfeldt, line 3: The `__future__` import would be made redundant by this (here and elsewhere).

Dockerfile
```diff
@@ -26,4 +26,4 @@ RUN apt-get update && \
             tidy \
             vim
 
-RUN pip install git+https://github.com/kattis/problemtools
+RUN pip3 install git+https://github.com/ghamerly/problemtools@move-to-py3
```
mpsijm, line 29: This line should probably be reverted before merging

examples/different/submissions/accepted/different.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env python2
```
mpsijm, line 1: This file could be renamed to `different_py2.py` to avoid confusion

mpsijm (commented):
This PR will most likely close #165 :smile:

Dockerfile
ghamerly: Fixed, thanks.

problemtools/verifyproblem.py
ghamerly: Fixed, thanks.

ghamerly:
I fixed the issues @mpsijm and @thorehusfeldt identified. I also made a few updates to the imports in the key scripts, and replaced optparse with argparse. I think this is ready for merging.

godmar:
I'm currently working on making a container that combines problem tools + tcardonne's github runner for github CI for naq20. If this PR's landing is imminent, I think I'd want to use it.

ghamerly:
> If this PR's landing is imminent, I think I'd want to use it.

Yes, I think we're going to merge it in the next day or so.

admin/docker/Dockerfile.icpc
```diff
@@ -28,15 +28,20 @@ RUN ln -s /usr/local/kotlinc/bin/* bin/
 
 # Reconfigure problemtools:
 # - Use PyPy for Python 2
+# - Use PyPy for Python 3
 # - Use /usr/local/bin rather than /usr/bin for Kotlin
 RUN mkdir -p /etc/kattis/problemtools
 RUN echo " \n\
 python2: \n\
```
RagnarGrootKoerkamp, line 35: Python2 is no longer supported at the worldfinals:
https://icpc.global/worldfinals/programming-environment

admin/docker/Dockerfile.icpc
```diff
@@ -28,15 +28,20 @@ RUN ln -s /usr/local/kotlinc/bin/* bin/
...
     name: 'Python 2 w/PyPy'\n\
     run: '/usr/bin/pypy \"{mainfile}\"'\n\
  \n\
+python3: \n\
+    name: 'Python 3 w/PyPy'\n\
+    run: '/usr/bin/pypy3 \"{mainfile}\"'\n\
+ \n\
```
RagnarGrootKoerkamp, line 42: Not sure how overriding works, but the shebang wouldn't be necessary anymore without python 2. (I'd love to just default to python3 by default without shebang, but in general this is bad for backwards compatibility.)

RagnarGrootKoerkamp (commented):
Using a different python interpreter for submissions in the ICPC image than in default settings is a bit weird, but making pypy3 the default is a bit annoying as well since not everybody has that installed by default.

admin/docker/Dockerfile.icpc
ghamerly: True, but it is possible that judges may still want to use it as a solution language.

ghamerly:
> Using a different python interpreter for submissions in the ICPC image than in default settings is a bit weird, but making pypy3 the default is a bit annoying as well since not everybody has that installed by default.

Sure. But pypy3 *is* in the docker image that this Dockerfile is building.

PR #181, opened 2020-12-05
Draft implementation of generators.yaml spec

SuprDewd:
This is a first attempt at an implementation of the `generators.yaml` draft spec described at
- https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/master/doc/generators.md
- https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/master/doc/generators.yaml

It adds the `generatedata` command which has the following synopsis:
```
usage: generatedata [-h] [-g] [-c] [-C] [-n] [-j PARALLELISM] [-b]
                    [-l LOG_LEVEL] [-e]
                    [--max_additional_info MAX_ADDITIONAL_INFO]
                    problemdir [problemdir ...]

Generate test data for a problem package in the Kattis problem format.

positional arguments:
  problemdir

optional arguments:
  -h, --help            show this help message and exit
  -g, --generate        generate test data
  -c, --clean           clean up generated files
  -C, --clean_all       clean up generated and unrecognized files
  -n, --dry_run         don't actually do anything
  -j PARALLELISM, --parallelism PARALLELISM
                        level of parallelism
  -b, --bail_on_error   bail verification on first error
  -l LOG_LEVEL, --log_level LOG_LEVEL
                        set log level (debug, info, warning, error, critical)
  -e, --werror          consider warnings as errors
  --max_additional_info MAX_ADDITIONAL_INFO
                        maximum number of lines of additional info (e.g.
                        compiler output or validator feedback) to display
                        about an error (set to 0 to disable additional info)
```

The `verifyproblem` command is also augmented to validate the corresponding `generators.yaml` file.

At the moment it does not support testgroup includes, and probably won't until their semantics have been fully decided. Other notable features that are also missing:
- Support for testdata.yaml
- First-class support for .interaction files
- Keep track of how files were generated for to determine what needs to be regenerated (in progress)


pehrsoderman:
@SuprDewd Can you rebase this, so we can get it moving forward?

SuprDewd:
@pehrsoderman I've finished rebasing.

@thorehusfeldt I added support for testdata.yaml. Do you want to assist with the review by taking this for a spin?

RagnarGrootKoerkamp:
@SuprDewd What exactly does `testdata.yaml` support mean in this case? Just allowing them to exist in the `data` directory? Or are you actually generating them, as I did here https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/108#issuecomment-826823499.

RagnarGrootKoerkamp:
If you'd like I can also go over the full code, but someone from the Kattis side should probably also do this.

problemtools/verifyproblem.py
```diff
@@ -740,6 +741,296 @@ def check(self, args):
...
+
+        seed = str(int(hashlib.sha512((random_salt + command).encode('utf-8')).hexdigest(), 16) % (2**31))
+
+        parts = shlex.split(command)
+        if not parts:
+            return err()
+
+        for i, part in enumerate(parts):
+            new = ''
+            for j, group in enumerate(part.split('{')):
```
RagnarGrootKoerkamp, line 796: Hmm, I'm just doing regex search replace in my code, so it also works for e.g. `seed={seed}`.
```
    SEED_REGEX = re.compile(r'\{seed(:[0-9]+)?\}')
    NAME_REGEX = re.compile(r'\{name\}')
```

problemtools/verifyproblem.py
```diff
@@ -740,6 +741,296 @@ def check(self, args):
...
+            self._testcases.append(state)
+
+    def _parse_directory(self, data, state):
+        # TODO: Process includes
+
+        if 'testdata.yaml' in data:
+            content = data['testdata.yaml']
+            if content is None:
+                content = {}
+            self._testdata_yaml['%s/%s' % (state['path'], 'testdata.yaml')] = content
```
RagnarGrootKoerkamp, line 834: These kind of things are much nicer using `pathlib.Path` IMO.

problemtools/verifyproblem.py
```diff
@@ -740,6 +741,296 @@ def check(self, args):
...
+                case_counter += 1
+
+            for name, value in sorted(case.items(), key=lambda kv: str(kv[0])):
+                if ordered:
+                    num = case_format % case_counter
+                    name = num + ('' if name is None else '-' + str(name))
+                else:
+                    name = str(name)
+
+                next_state = copy.deepcopy(state)
```
RagnarGrootKoerkamp, line 859: Doing a deepcopy doesn't feel super nice.

problemtools/verifyproblem.py
```diff
@@ -740,6 +741,296 @@ def check(self, args):
...
+                next_state['path'] = '%s/%s' % (state['path'], name)
+                self._parse_element(value, next_state)
+
+    def _parse_element(self, data, state):
+        if sys.version_info[0] < 3 and data == '':
+            data = None
+        if data is None:
+            data = '/%s.in' % state['path']
```
RagnarGrootKoerkamp, line 867: Are you specifically not using f-strings for python < 3.7 compatibility?

problemtools/verifyproblem.py
```diff
@@ -740,6 +741,296 @@ def check(self, args):
...
+        if not isinstance(data, dict):
+            self.error("Path %s in generators.yaml must specify a dict" % state['path'])
+            return
+
+        state.update({
+            key: data[key]
+            for key in Generators._TESTCASE_OPTIONS
+            if key in data
+        })
+        if sys.version_info[0] < 3:
```
RagnarGrootKoerkamp, line 880: Can you add a comment what this if is doing?

problemtools/generatedata.py
```diff
@@ -0,0 +1,272 @@
...
+    args = None
+
+def generate_case(case_idx):
+    ProblemAspect.errors = 0
+    ProblemAspect.warnings = 0
+    prob = GenerateState.prob
+    args = GenerateState.args
+    case = prob.generators._testcases[case_idx]
+
+    steps = [
```
RagnarGrootKoerkamp, line 118: oh this is nice! Maybe I can steal this idea to simplify some code ;)

RagnarGrootKoerkamp (approved):
Overall this looks much cleaner, simpler, and shorter than my own implementation, so good job :+1:

problemtools/generatedata.py
RagnarGrootKoerkamp: Do you run input/output validators on the generated cases?

problemtools/verifyproblem.py
SuprDewd: Right, and I believe the current implementation also allows that. If I recall correctly the motivation for doing it this way was to:
1) Detect invalid expressions, e.g. {undefined} or {{seed, which the regex presumably would not.
2) Make sure arguments are grouped appropriately when passed on to the generator. E.g. in the case of --infile {name}.in, where the name of the case is something like "hello world" (which admittedly is not allowed by the spec), then this implementation would pass ["--infile", "hello world.in"] while a regex-based one would pass ["--infile", "hello", "world.in"].

RagnarGrootKoerkamp: Ah I see; yes the code makes sense on closer inspection :+1:

problemtools/verifyproblem.py
SuprDewd: I identify particular entries within the generators.yaml file with a slash-separated sequence of keys, e.g. `data: sample: solution: test.cpp` is identified with `data/sample/solution`. This is then used in error messages and other places. While this may in many cases correspond to a particular file, it won't always, so it felt wrong to be using `pathlib` here.

problemtools/verifyproblem.py
SuprDewd: Did you have a cleaner alternative in mind?

problemtools/verifyproblem.py
SuprDewd: Yes. As far as I know problemtools runs even on Python 2, so I wanted to maintain the same level of backwards compatibility.

problemtools/verifyproblem.py
SuprDewd: This was accounting for some odd behavior in the python2 yaml library, but it seems like it has been fixed as I cannot reproduce it anymore. Removed.

problemtools/generatedata.py
SuprDewd: No, that's left as the responsibility of the `verifyproblem` command.

SuprDewd:
Can we get this merged? This has been tested at three NWERCs without issues.

RagnarGrootKoerkamp:
Maybe we should also merge the spec itself:) One thing I was thinking of that may be nice would be a way to directly write small testcases in the generators.yaml, avoiding the need for an `stsout.py` generator or separate files.

PR #182, opened 2020-12-15
Fix bug in guess output validator

RagnarGrootKoerkamp:
Currently it's possible to have the following interaction with the adaptive interactor:
```
400
higher
700
higher
1
higher
600
lower
```

We ran into this because we increased the number of allowed queries to `1000` for the BAPC testsession and somebody submitted a program guessing `37*i%1000+1` on the `i`th try.

### DIFF
--- a/examples/guess/output_validators/guess_validator/validate.cc
+++ b/examples/guess/output_validators/guess_validator/validate.cc
@@ -62,11 +62,11 @@ void check_case() {
         } else if (diff < 0) {
             cout << "lower\n";
             cout.flush();
-            sol_hi = guess-1;
+            sol_hi = min(sol_hi, guess-1);
         } else {
             cout << "higher\n";
             cout.flush();
-            sol_lo = guess+1;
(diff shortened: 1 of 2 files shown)


pehrsoderman:
@RagnarGrootKoerkamp Can you add an example solution triggering this bug to the problem package as well?

RagnarGrootKoerkamp:
@pehrsoderman done, but it's not so effective: the solution needs 1000 queries to be AC, but the limit is only 10 here so it will be WA anyways.

pehrsoderman:
@RagnarGrootKoerkamp Can you add something that, for example, makes the solution crash out if the bug is present? And a comment, to make it a little less 😲 if you don't know what is there for?

Sorry for the nitpicking. I should have been more clear to start with.

RagnarGrootKoerkamp:
@pehrsoderman done now! The submission should get WA, but currently gets an RTE.

pehrsoderman:
@RagnarGrootKoerkamp Great, thanks!

PR #192, opened 2021-03-17
Start allowing input_validators instead of input_format_validators per the spec.

jsannemo:


RagnarGrootKoerkamp:
See also #185, but this one is probably a bit nicer; I like the deprecation warning :+1:

jsannemo:
> See also #185, but this one is probably a bit nicer; I like the deprecation warning

Twice the PRs surely means twice the chance of getting it in!

PR #195, opened 2021-04-30
PvP problem support for problemtools

default-username-852:
This PR extends problemtools to support validation of PvP problems as defined by https://github.com/Kattis/problem-package-format/pull/12.

As with the extension to the problem package format, this was developed by 8 students as part of a course at KTH which has finished. If someone wants to keep working on this, feel free to take what we have done and keep on building on it!


niemela:
Sadly this will not be supported any time soon.

PR #198, opened 2021-06-08
First draft of more fine-grained logging

thorehusfeldt:
Moves most of the functionality from ProblemAspect
into various filters, loggers, and handlers in logger.py.

Why is this good?
- closer to how python's own logging facility handles logging (https://docs.python.org/3/library/logging.html).
- allows more flexible and fine-grained logging, in particular 
    - easy to add a file handler that sends DEBUG or INFO log to `/tmp`
    - attractive and clean way to add alternative formatting or more fine-grained logs (say, judgemessages or graders). This is the _original_ motivation behind this refactoring.
- decreases dependencies in verifyproblem's class structure (ideally, would get rid of `ProblemAspect`) by decoupling logging (`log.error`) from problem parts/aspects (`self.get_score_range()`)

Why is this bad?
- considerable refactoring of code

Stuff to think about before moving on:
- Is this worth doing at all?
- Can the attribute `ProblemAspect._check_res` be moved to `InputValidators`? Seems to be used only there.
- How fine-grained should the logging hierarchy be? Currently the leaves follow the same granularity as does `ProblemAspect` (for instance, `verifyproblem.hello.submissions`), but one could go all the way down to test case or submission level `verifyproblem.hello.test case group.group 3/045-huge.in` or `verifyproblem.hello.submissions.ac.th.py`.

### DIFF
--- a/problemtools/logger.py
+++ b/problemtools/logger.py
@@ -1,0 +1,308 @@
+"""
+Logging for problemtools.verifyproblem.
+
+The logging hierarchy has depth two and uses different logger classes.
+Every problem is at the top of a logging hierarchy, below which are the problem
+aspects (which may be parts such # as "output validators" or test cases):
+
+  verifyproblem: RootLogger
+  |
+  +- hello: ProblemLogger
+  |     |
+  |     +- submissions: AspectLogger
+  |     +- test case secret/002-huge.in: AspectLogger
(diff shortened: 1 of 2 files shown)


pehrsoderman:
@thorehusfeldt This needs a rebase.

PR #209, opened 2022-07-12
Improved logging, fix unicode error, include_dir buildrun

### LINKED ISSUE #201: UnicodeDecodeError problems due to non-parseable _JUNK_CASES
When validating problems with the current develop branch, I get the following issue:

`UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 137: invalid start byte`, thrown in `verifyproblem.py", line 1141, in __get_feedback:  all_feedback.append(feedback.read(128*1024))`

The problem is that the validator (based on `validate.h`) reports
```
TC 1: Judge says possible but author says !"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~��������������������������������������������������������������������������������������������������������������������������������!
```
(output of `cat judgemessage.txt`).

This in turn is caused by the following entry in `_JUNK_CASES`: `('a binary file with byte values 0 up to 256', bytearray(x for x in range(256)))`. This creates an answer which contains non-unicode characters, which then fails to parse because when reading feedback, `verifyproblem.py` expects unicode answers.

Commenting out this line fixes the problem. I suggest one should either use every printable character / random unicode characters or make the parsing of judgemessage more robust (iterate character by character over the input and if it cannot be parsed replace with questionmark or similar?)

incaseoftrouble:
Hi,

I'm using problemtools as library in a larger script to automate deployment of lectures. In that process I found some things I thought needed fixing. I tried to keep changes small and can also make them a bit more consistent. Depending on your wishes, I might address some more things in the future :-)

Mainly, I introduced a bit more consistent logging and added the possibility of having `build/run` in the include_dir. I used this for some java validators where a common build script lies in an included dir.

Closes #201

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -24,6 +24,8 @@ from . import config
 from . import languages
 from . import run
 
+log = logging.getLogger(__name__)
+
 
 def is_TLE(status, may_signal_with_usr1=False):
     return (os.WIFSIGNALED(status) and
@@ -82,6 +84,7 @@ class ProblemAspect:
     errors = 0
     warnings = 0
     bail_on_error = False
+    consider_warnings_errors = False
     _check_res = None
     basename_regex = re.compile('^[a-zA-Z0-9][a-zA-Z0-9_.-]*[a-zA-Z0-9]$')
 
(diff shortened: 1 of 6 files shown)


problemtools/verifyproblem.py
```diff
@@ -234,26 +239,31 @@ def _run_submission_real(self, sub, args, timelim_low, timelim_high):
...
+                    with open(errfile, mode="rt") as f:
+                        info = f.read()
+                res2 = SubmissionResult('RTE', additional_info=info)
             else:
                 res2 = self._problem.output_validators.validate(self, outfile)
             res2.runtime = runtime
-        if sys.stdout.isatty():
-            sys.stdout.write('%s' % '\b \b' * (len(msg)))
+        # if sys.stdout.isatty():
+        #     sys.stdout.write('%s' % '\b \b' * (len(msg)))
```
pehrsoderman, line 266: Why is this commented out?

pehrsoderman (changes requested):
There seems to be some debugging related changes left in this pr...

incaseoftrouble:
Hey, sorry, I somehow completely forgot that this is open. Should I update this to the current release?

pehrsoderman:
@incaseoftrouble Please do a rebase

incaseoftrouble:
Just as a remark on that: My schedule is absolutely full, I put this into my calendar for February. (But I put it in!)

incaseoftrouble:
Moved to #253

PR #215, opened 2023-01-06
Do not grade empty groups

### LINKED ISSUE #189: Default grader now crashes on empty test groups with mode min
Commit 4251179bd16087f9ba09a327bff274600b2ce180 based on pull request #174 changed the behaviour of the default grader in certain corner cases from `JE 0` to `JE`,see

https://github.com/Kattis/problemtools/commit/4251179bd16087f9ba09a327bff274600b2ce180#commitcomment-46068430

This leads to `verifyproblem` with the `-d data` flag crashing in normal use:

```
> verifyproblem . -d a
[...]
File "/usr/lib/python3/dist-packages/problemtools/verifyproblem.py", line 567, in aggregate_results
    if not (min_score <= score <= max_score) and not self._seen_oob_scores:
TypeError: '<=' not supported between instances of 'float' and 'NoneType'
```

Indeed, `min_score` is typically `0.0`, and `score` is now `None` (rather than `0`).

## Why does this happen?

A scoring problem with a grader aggregation mode `min` wants to compute the total score of an empty test group. 

(The empty test group occurs during problem development because the `-d` flag to `verifyproblem` was used to only run the submissions on a certain test group, such as `-d group3`, or on certain input files not occurring in all test groups, such as `-d huge`. This flag _correctly_ leads to empty test groups.)

The behaviour of `min` on aggregating the empty set is undefined in the specification, but the choice `JE 0`, though producing the unhelpful verdict `JE` also produced the score `0`, making the behaviour confusing but useful during problem development.

With the choice `JE`, verifyproblem’s verdict is still the unhelpful `JE`, but now the variable `score` is set to None, leading to a run-time error in the comparison in line 567 quoted above and aborting verification.

 ## Fix (quick)
 
 Go back to `JE 0` instead of `JE`.
 
 ## Fix (better)
 
 Discuss what `min([])` should mean. The most useful behaviour would be `AC 0` or even verdict `AC` and score `accept_score`, but with a warning “there were empty test groups.”
 
 ## Minimal example

Here is a baby problem about adding two integers with two secret input files `a.in` and `b.in`, each in their own test group.
 
To crash:
```
verifyproblem . -d a
```
[JE.zip](https://github.com/Kattis/problemtools/files/5829970/JE.zip)

ghamerly:
Fixes #189

This causes verifyproblem to give a verdict of `AC` with score of `0` to any empty group, rather than calling the grader on an empty group. It also changes the way the default_grader reports a judge error (reporting a score as well).

These changes are mostly only useful for use in verifyproblem with the `-d` flag, which can filter out test files so that groups appear empty.

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -497,6 +497,9 @@ class TestCaseGroup(ProblemAspect):
             if not f[:-4] + '.in' in infiles:
                 self.error("No matching input file for answer '%s'" % f)
 
+        if not self.get_subgroups() and not self.get_testcases():
+            self.error('Test case group is empty')
+
         # Check whether a <= b according to a natural sorting where numeric components
         # are compactified, so that e.g. "a" < "a1" < "a2" < "a10" = "a010" < "a10a".
         def natural_sort_le(a, b):
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -1320,6 +1320,10 @@ class Graders(ProblemAspect):
         verdict = 'AC'
         score = 0
 
+        if not sub_results:
+            self.info('No results on %s, so no graders ran' % (testcasegroup,))
+            return (verdict, score)
+
         grader_flags = testcasegroup.config['grader_flags'].split()
         self.debug('Grading %d results:\n%s' % (len(sub_results), grader_input))
         self.debug('Grader flags: %s' % grader_flags)
--- a/support/default_grader/default_grader
+++ b/support/default_grader/default_grader
@@ -68,4 +68,4 @@ try:
     score = aggregate_scores(scores)
     print('%s %f' % (verdict, score))
 except:
-    print('JE')
+    print('JE 0')


niemela:
Could this ever happen without using `-d`? I.e. is there something explicitly disallowing empty groups? If not, it might be "more correct" to have the verdict be AC with `accept_score` instead?

ghamerly:
@niemela There was previously nothing in verifyproblem that reported a warning or an error if a test case group was empty (actually empty, not just filtered out with `-d`). With the commit I just added (4e6179e), I have made that situation an error. If you think it's better treated as a warning, I'm open to that.

As far as using `accept_score` instead of `0` as the default score (when a test group is empty, due to filtering), we can do that easily (I have written but not committed the code that would do that). However, I'm not sure that the semantics are quite correct to apply it this way, because `accept_score` is used on each test *case*, not on each test *group*. Thoughts?

pehrsoderman (approved):

PR #216, opened 2023-01-08
Updates for plastex 3.0

### LINKED ISSUE #214: Required plasTeX version?
Does ProblemTools need an older version of plastTeX as a dependency? I get the following error when I try to run `problem2html`:

```
Traceback (most recent call last):
  File "/u/evouga/anaconda3/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/u/evouga/anaconda3/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/u/evouga/icpc/problemtools/problemtools/problem2html.py", line 140, in <module>
    main()
  File "/u/evouga/icpc/problemtools/problemtools/problem2html.py", line 136, in main
    convert(options.problem, options)
  File "/u/evouga/icpc/problemtools/problemtools/problem2html.py", line 41, in convert
    tex = plasTeX.TeX.TeX(myfile=texfile)
TypeError: __init__() got an unexpected keyword argument 'myfile'
```

ghamerly:
Fixes #214 

I have done a bit of testing and this seems to work. This tries to maintain backwards compatibility with the old plasTeX 2.1 while supporting the newer plasTeX 3.0.

As of this writing, the debian package `python3-plastex` gives the old version of plasTeX (2.1), so to test this with the new plasTeX, you should install via pip:
```
python3 -m pip install plastex==3.0
```
and then running bin/problem2html.sh (or however you choose to run it, e.g. via an installed debian package). If you want to test it with the debian package, I had to install it with `dpkg --force-all -i kattis_problemtools*.deb` to install it without the  `python3-plastex` dependency already installed.

### DIFF
--- a/problemtools/problem2html.py
+++ b/problemtools/problem2html.py
@@ -38,7 +38,7 @@ def convert(problem, options=None):
 
         # Setup parser and renderer etc
 
-        tex = plasTeX.TeX.TeX(myfile=texfile)
+        tex = plasTeX.TeX.TeX(file=texfile)
 
         ProblemsetMacros.init(tex)
 
--- a/problemtools/problem2html.py
+++ b/problemtools/problem2html.py
@@ -38,7 +38,12 @@ def convert(problem, options=None):
 
         # Setup parser and renderer etc
 
-        tex = plasTeX.TeX.TeX(file=texfile)
+        try:
+            # try the new version of plasTeX (in version 3.0, the argument is named "file")
+            tex = plasTeX.TeX.TeX(file=texfile)
+        except:
+            # fall back to the old version of plasTeX (previous versions used "myfile")
+            tex = plasTeX.TeX.TeX(myfile=texfile)
 
         ProblemsetMacros.init(tex)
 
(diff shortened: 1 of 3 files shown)


problemtools/problem2html.py
```diff
@@ -38,7 +38,12 @@ def convert(problem, options=None):
 
         # Setup parser and renderer etc
 
-        tex = plasTeX.TeX.TeX(myfile=texfile)
+        try:
+            # try the new version of plasTeX (in version 3.0, the argument is named "file")
+            tex = plasTeX.TeX.TeX(file=texfile)
```
pehrsoderman, line 43: This looks rather fugly to me. Could you peek at plasTeX.__version__ instead of the try and bare except dance?

pehrsoderman (changes requested):

problemtools/problem2html.py
ghamerly: Your suggestion was my first idea as well, but unfortunately right now the plasTeX source has the incorrect `__version__`, which I have reported here: https://github.com/plastex/plastex/issues/308

So for now (until that is fixed) there's no good way to tell which version of plasTeX you have without either doing this method (try/except) or trying some inspection on the name of the arguments, which also felt ugly.

pehrsoderman: @ghamerly PlasTeX were rather responsive, and your issue seems to be fixed already ;)

ghamerly: Good; I'll push a better fix now that the version in plasTeX is fixed.

pehrsoderman (approved):

PR #222, opened 2023-03-04
Handle plastex version diffs by try-catch instead of checking __version__

### LINKED ISSUE #219: problem2html crashing
I'm getting the following error message when running `problem2html` after updating problemtools:
```
Traceback (most recent call last):
  File "/usr/local/bin/problem2html", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.8/dist-packages/problemtools/problem2html.py", line 140, in main
    convert(options.problem, options)
  File "/usr/local/lib/python3.8/dist-packages/problemtools/problem2html.py", line 43, in convert
    tex = plasTeX.TeX.TeX(**{argname: texfile})
TypeError: __init__() got an unexpected keyword argument 'myfile'
```
I've never seen this before and it happens with all problem directories, including past contest problems that used to compile perfectly into HTML. Within `verifyproblem`, this shows up as
```
ERROR in problem statement: Could not convert problem statement to html for language "".  Run problem2html  on the problem to diagnose.
```
I have plasTeX version 2.1, and this problem likely comes from 5208a609fff4c2f8541c6f639a4a414f20dfed11

simonlindholm:
Fixes #219. Plastex has gone 2 months now without a new release, and this is actively hurting people using problemtools. Urging plastex to make a new release won't help fully either, because some people will still be stuck on the old plastex version, and it's not immediately clear that you need to `python3 -m pip install --upgrade plastex`.

### DIFF
--- a/problemtools/problem2html.py
+++ b/problemtools/problem2html.py
@@ -38,9 +38,13 @@ def convert(problem, options=None):
 
         # Setup parser and renderer etc
 
-        # plasTeX version 3 changed the name of this argument
-        argname = 'myfile' if float(plasTeX.__version__) < 3 else 'file'
-        tex = plasTeX.TeX.TeX(**{argname: texfile})
+        # plasTeX version 3 changed the name of this argument (and guarding against this
+        # by checking plasTeX.__version__ fails on plastex v3.0 which failed to update
+        # __version__)
+        try:
+            tex = plasTeX.TeX.TeX(myfile=texfile)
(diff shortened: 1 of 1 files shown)


jsannemo:
Yes please!!!!!

ghamerly (approved):
I approve; this was the approach I originally proposed to fix the v3 argument name issue, but people didn't like it, so I went with the `__version__` check... 😄

PR #233, opened 2023-07-09
Basic support for (programming) languages in problem.yaml

Tagl:
I needed this functionality for a problem and noticed the not supported error.
Definitely needs some polishing.

This PR adds support for filtering supported languages by specifying them in problem.yaml

### DIFF
--- a/problemtools/config/problem.yaml
+++ b/problemtools/config/problem.yaml
@@ -25,6 +25,7 @@ grading:
    objective: max
    show_test_data_groups: False
 
+languages: all
+
 # These are in the spec but currently unsupported
 libraries: ''
-languages: ''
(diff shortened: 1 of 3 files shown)


RagnarGrootKoerkamp:
Oh we also just added this for BAPCtools, RagnarGrootKoerkamp/bapctools/pull/256. We use the name field in problem.yaml for this (as per the spec). Can you explain what your implementation does?

Tagl:
It simply creates a new Languages object that stores the supported programming languages, and replaces the global config with that before running all the checks.
The new config will only include the languages with the specified language ids.
If a language id is specified but does not exist in the global language config then an error is reported.

I notice now that the default should be `all`, so will have to change that.

RagnarGrootKoerkamp:
Oh this is about programming languages, not problem statement languages. nevermind my previous comment then.

PR #240, opened 2023-10-09
Add python type annotations to `problem2{html,pdf}` and `verifyproblem`

JoelNiemela:
Adds type annotations to `problem2html.py`, `problem2pdf.py`, and `verifyproblem.py`.
Also modifies those files slightly, but the vast majority of changes are type annotations.

### DIFF
--- a/problemtools/generatedata.py
+++ b/problemtools/generatedata.py
@@ -13,7 +13,7 @@ from .verifyproblem import Generators, ProblemAspect, Problem, is_RTE, argparser
 
 ALL_EXTENSIONS = ['in', 'ans'] + Generators._VISUALIZER_EXTENSIONS
 
-def argparser():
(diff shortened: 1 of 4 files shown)


JoelNiemela:
I would personally ignore the CodeFactor errors here, so I'll leave it as-is unless anyone else has a contrary opinion.

Tagl (approved):
Looks mostly good, however

JoelNiemela:
> > I would personally ignore the CodeFactor errors here, so I'll leave it as-is unless anyone else has a contrary opinion.
> 
> Instead of ignoring the warning that the code is breaking the currently defined rules, either fix it to follow the rules or suggest how the rules should be redefined.

I've had people disagree with me about this specific thing recently, so I think my opinion is more "controversial" than I thought, so I'll just change it to fix the CodeFactor errors.

PR #253, opened 2024-02-12
Improved logging, fix unicode error, include_dir buildrun

### LINKED ISSUE #201: UnicodeDecodeError problems due to non-parseable _JUNK_CASES

incaseoftrouble:
This replaces #209  (I migrated the repository) and fixes #201


incaseoftrouble:
Okay, I'm quite happy with this now.

Changes:
 * Use (named) loggers where applicable, in particular each problem aspect now has their own name
 * Building can now include files (e.g. if you have a common build script to build a java-based tool, you don't need to copy it around)
 * Fixed a problem with _JUNK_CASES: Bytes larger than 127 are not valid UTF-8, so validators could fail with UTF8DecodeError
 * Add a bit more logging here and there

problemtools/verifyproblem.py
```diff
@@ -1165,7 +1186,7 @@ def __str__(self) -> str:
 
 _JUNK_CASES = [
     ('an empty file', b''),
-    ('a binary file with byte values 0 up to 256', bytearray(x for x in range(256))),
+    ('a binary file with byte values 0 up to 127', bytearray(x for x in range(127))),
```
simonlindholm, line 1189: I don't think this is the right fix -- output validators that crash on invalid UTF-8 is a potential source of judge errors that we want to catch. Better change how judge messages are read to cope with invalid UTF-8.

incaseoftrouble: Ok thinking about it, it makes sense to require that every validator should be able to cope with any garbage input.

I'll try to propose a change, give me a moment.

incaseoftrouble: I replaced this with 1024 "random" bytes (generated with a fixed seed) and fixed verifyproblem crashing due to non-unicode problems.

problemtools/problem2html.py
```diff
@@ -7,14 +7,15 @@
...
 
-import plasTeX.TeX
-import plasTeX.Logging
-
-from .ProblemPlasTeX import ProblemRenderer
-from .ProblemPlasTeX import ProblemsetMacros
-from . import template
 
 def convert(options: argparse.Namespace) -> None:
+    import plasTeX.TeX
```
pehrsoderman, line 12: Unless we have a strong reason not to, we try to follow PEP8. In PEP8 imports should be at the top of the file.

pehrsoderman (changes requested):
Minor comment regarding pep8.

problemtools/problem2html.py
incaseoftrouble: Hmm right I should have moved that to a separate issue. The intent was that verifyproblem then runs without plastex installed.

But I don't have a strong opinion on this, can revert this.

incaseoftrouble: fine with me. But then let's be honest about the plastex dependency and add it to requirements.txt :-) (added)

incaseoftrouble: Nvm, it is in setup.py!

pehrsoderman (approved):

simonlindholm:
> Use (named) loggers where applicable, in particular each problem aspect now has their own name

I've found that this makes the output look rather uglier. Where previously I had
```
$ verifyproblem.sh .
Loading problem myproblemid
Checking config
...
Checking submissions
ERROR in submissions: Require at least one "accepted" submission
   Slowest AC runtime: None, setting timelim to 300 secs, safety margin to 300 secs
```
I now get
```
$ verifyproblem.sh .
Loading problem myproblemid
Checking config
...
Checking submissions
[__main__.myproblemid.submission] ERROR: Require at least one
   "accepted" submission
   Slowest AC runtime: None, setting timelim to 300 secs, safety margin to 300 secs
```

I would prefer to revert the use of `[]`-style log prefixes here, because this is user-facing UI and not logging that needs to be made systematic. But at the very least we should remove the mention of `__main__` and avoid 80-col text wrapping.

Tagl:
I 100% agree with the text wrapping being worse now than previously. I do see great value in categorization of the log messages but I'd also prefer it to not look this way when running verifyproblem in default log level outputting to stdout.

incaseoftrouble:
> []-style log prefixes ... avoid 80-col text wrapping.

This is due to PlasTeX completely butchering the logging framework in a (near) irreparable way. Which is why I wanted to delay loading it.

> I would prefer to revert the use of []-style log prefixes here, because this is user-facing UI and not logging that needs to be made systematic. But at the very least we should remove the mention of __main__

The correct "fix" for this is to remove eagerly importing plastex and, if needed, define a logging format that is applied in `__main__`. Then the output will look like before with the added bonus that problemtools is usable as a library.

Issue #254, opened 2024-02-15
verifiy problem not working


EinarDarri:
![image](https://github.com/Kattis/problemtools/assets/54663445/53d91767-c2fe-492b-b4b4-b8c7736c6ab1)

I just installed problemtools and it seams that the latest merge #253  added an error

this error is showing up on all problems that I try to verify

incaseoftrouble:
Hi, this is my fault, I merged something and forgot to check. You can use #256 until it is merged

gkreitz:
#256 is merged, closing.

PR #255, opened 2024-02-15
Fix run time errors from logging changes in #253

Tagl:
The latest merge causes verifyproblem to crash.
These are minimal changes required to prevent that.

Closes #254

PR #256, opened 2024-02-18
Fix missed init, add test case

incaseoftrouble:
Hi,

I saw I missed one `super().__init__`, sorry! It is fixed now and I added a (shallow) test to check that initialization works.

This doesn't execute a full verifyproblem call (which I should have done before committing!) since pytest and os.fork apparently don't get along very well.

(Should `program.py` be migrated to a subprocess call instead? Setting limits can be done be done with `preexec_fn`)

### DIFF
--- a/problemtools/tests/test_verify_hello.py
+++ b/problemtools/tests/test_verify_hello.py
@@ -1,0 +1,17 @@
+import pathlib
+import problemtools.verifyproblem as verify
+
+
+def test_load_hello():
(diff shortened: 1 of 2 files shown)


Tagl:
Like #255 this would close #254

SvartaHjarta:
AUR package remains broken because of this, is there anything I can do to help this along?

pehrsoderman (approved):

PR #257, opened 2024-02-22
Explicitly set language versions for Java and Kotlin in languages.yaml

mpsijm:
Explicitly setting the language version in the compiler options makes it easier to locally debug compile errors when running `verifyproblem` outside Docker. Note that the Kotlin version `1.3` will need to be upgraded to `1.8` soon (see #246), but @ghamerly told me that `1.3` is currently what's still used for installing the problems on Kattis.

It would be nice to also have something similar for Python, but `python` does not have a flag for setting the language version. This would require to explicitly refer to `/usr/bin/python3.8`, but that means I'd need to explicitly install Python 3.8 next to 3.11 on my system, which I'd rather avoid :stuck_out_tongue:

### DIFF
--- a/problemtools/config/languages.yaml
+++ b/problemtools/config/languages.yaml
@@ -147,7 +147,7 @@ java:
     name: 'Java'
     priority: 800
     files: '*.java'
-    compile: '/usr/bin/javac -encoding UTF-8 -sourcepath {path} -d {path} {files}'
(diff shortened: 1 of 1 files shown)


pehrsoderman:
Pinning language versions like this is a bit of a maintenance headache. I am not sure if the gain is worth the cost.

pehrsoderman (approved):
After some thinking we will merge this.

PR #261, opened 2024-05-24
Check that all symlinks point to something existing within the problem package

gkreitz:
We recently saw issues caused by dangling symlinks being present in a problem package. This PR makes it an error to have any symlinks anywhere in the problem package point to something which does not exist, or point somewhere outside of the problem package.

Example output:
```
Loading problem hello
ERROR in hello: Symlink submissions/foo links to /home which is outside of problem package
ERROR in hello: Symlink submissions/foo links to /home which is an absolute path. Symlinks must be relative.
ERROR in hello: Symlink submissions/accepted/hello2.cc links to helloooooo.cc which does not exist
ERROR in hello: Symlink data/oops links to ../../data/sample which does not exist
ERROR in hello: Symlink data/oops links to ../../data/sample which is outside of problem package
...
```

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -1851,6 +1851,8 @@ class Problem(ProblemAspect):
             if not re.match('^[a-z0-9]+$', self.shortname):
                 self.error(f"Invalid shortname '{self.shortname}' (must be [a-z0-9]+)")
 
+            self._check_symlinks()
+
             run.limit.check_limit_capabilities(self)
 
             for part in args.parts:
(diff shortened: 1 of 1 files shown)


niemela:
Should we mention this in the spec? Or is it *obvious* that symlinks must point to something and that it mustn't point outside of the package?

niemela:
> It's not obvious that symlinks are allowed

That's a very good point. (And the rest follows).

Tagl:
Well do we want to allow symlinks to point to absolute paths? That seems
not so great me

gkreitz:
> Well do we want to allow symlinks to point to absolute paths? That seems not so great me

Excellent point. Absolute symlinks, even if they point to the right place on the machine we're currently running on feel like a very bad idea in a problem package. I added another check for absolute symlinks.

pehrsoderman (approved):

PR #263, opened 2024-06-06
Add -j flag for multi-threaded validation

simonlindholm:
Only parallelizes test case execution and submission compilation at the moment; input validation could be added with a bit of additional work, but it felt less important to me since it takes almost no time with validators written in C++ as we usually have it.

mypy is happy with the refactoring and I've done some lightweight testing, but it's possible that there's some breakage along less happy paths, though probably restricted to -j.

### DIFF
--- a/problemtools/run/__init__.py
+++ b/problemtools/run/__init__.py
@@ -16,7 +16,7 @@ from . import rutil
 
 
 def find_programs(path, pattern='.*', language_config=None, work_dir=None,
-                  include_dir=None, allow_validation_script=False):
+                  include_dir=None, allow_validation_script=False) -> list[Program]:
     """Find all programs in a directory.
 
     Args:
@@ -62,7 +62,7 @@ def find_programs(path, pattern='.*', language_config=None, work_dir=None,
 
 
 def get_program(path, language_config=None, work_dir=None, include_dir=None,
-                allow_validation_script=False):
+                allow_validation_script=False) -> Program|None:
     """Get a Program object for a program
 
     Args:
--- a/problemtools/run/buildrun.py
+++ b/problemtools/run/buildrun.py
@@ -27,6 +27,8 @@ class BuildRun(Program):
             work_dir (str): name of temp directory in which to run the
                 scripts (if None, will make new temp directory).
         """
+        super().__init__()
+
         if not os.path.isdir(path):
             raise ProgramError('%s is not a directory' % path)
 
@@ -53,32 +55,27 @@ class BuildRun(Program):
         if not os.access(build, os.X_OK):
             raise ProgramError('%s/build is not executable' % path)
 
-    def __str__(self):
+    def __str__(self) -> str:
         """String representation"""
         return '%s/' % (self.path)
 
 
-    _compile_result = None
-    def compile(self):
+    def do_compile(self) -> tuple[bool, str|None]:
         """Run the build script."""
-        if self._compile_result is not None:
-            return self._compile_result
-
         with open(os.devnull, 'w') as devnull:
             status = subprocess.call(['./build'], stdout=devnull, stderr=devnull, cwd=self.path)
         run = os.path.join(self.path, 'run')
 
         if status:
-            log.debug('Build script failed (status %d) when compiling %s', status, self.name)
-            self._compile_result = (False, f'build script failed with exit code {status:d}')
+            logging.debug('Build script failed (status %d) when compiling %s\n', status, self.name)
+            return (False, 'build script failed with exit code %d' % (status))
         elif not os.path.isfile(run) or not os.access(run, os.X_OK):
-            self._compile_result = (False, 'build script did not produce an executable called "run"')
+            return (False, 'build script did not produce an executable called "run"')
         else:
-            self._compile_result = (True, None)
-        return self._compile_result
+            return (True, None)
 
 
-    def get_runcmd(self, cwd=None, memlim=None):
+    def get_runcmd(self, cwd=None, memlim=None) -> list[str]:
         """Run command for the program.
 
         Args:
@@ -89,6 +86,6 @@ class BuildRun(Program):
         return [os.path.join(path, 'run')]
 
 
-    def should_skip_memory_rlimit(self):
+    def should_skip_memory_rlimit(self) -> bool:
         """Ugly hack (see program.py for details)."""
         return True
--- a/problemtools/run/checktestdata.py
+++ b/problemtools/run/checktestdata.py
@@ -22,27 +22,23 @@ class Checktestdata(Executable):
         if Checktestdata._CTD_PATH is None:
             raise ProgramError(
                 'Could not locate the Checktestdata program to run %s' % path)
-        super(Checktestdata, self).__init__(Checktestdata._CTD_PATH,
-                                            args=[path])
+        super().__init__(Checktestdata._CTD_PATH, args=[path])
 
 
-    def __str__(self):
+    def __str__(self) -> str:
         """String representation"""
         return '%s' % (self.args[0])
 
 
-    _compile_result = None
-    def compile(self):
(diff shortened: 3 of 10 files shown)


problemtools/run/__init__.py
```diff
@@ -62,7 +62,7 @@ def find_programs(path, pattern='.*', language_config=None, work_dir=None,
 
 
 def get_program(path, language_config=None, work_dir=None, include_dir=None,
-                allow_validation_script=False):
+                allow_validation_script=False) -> Program|None:
```
Tagl, line 65: Perhaps use
```suggestion
                allow_validation_script=False) -> Optional[Program]:
```
instead?

problemtools/run/buildrun.py
```diff
@@ -53,32 +55,27 @@ def __init__(self, path, work_dir=None, include_dir=None):
...
 
-    def __str__(self):
+    def __str__(self) -> str:
         """String representation"""
         return '%s/' % (self.path)
 
 
-    _compile_result = None
-    def compile(self):
+    def do_compile(self) -> tuple[bool, str|None]:
```
Tagl, line 63:
```suggestion
    def do_compile(self) -> tuple[bool, Optional[str]]:
```

problemtools/run/checktestdata.py
```diff
@@ -22,27 +22,23 @@ def __init__(self, path):
...
 
-    def __str__(self):
+    def __str__(self) -> str:
         """String representation"""
         return '%s' % (self.args[0])
 
 
-    _compile_result = None
-    def compile(self):
+    def do_compile(self) -> tuple[bool, str|None]:
```
Tagl, line 33:
```suggestion
    def do_compile(self) -> tuple[bool, Optional[str]]:
```

problemtools/run/program.py
```diff
@@ -50,12 +55,23 @@ def run(self, infile='/dev/null', outfile='/dev/null', errfile='/dev/null',
 
         return status, runtime
 
-    def code_size(self):
+    def compile(self) -> tuple[bool, str|None]:
```
Tagl, line 58:
```suggestion
    def compile(self) -> tuple[bool, Optional[str]]:
```

problemtools/run/program.py
```diff
@@ -50,12 +55,23 @@ def run(self, infile='/dev/null', outfile='/dev/null', errfile='/dev/null',
...
+        with self._compile_lock:
+            if self._compile_result is None:
+                self._compile_result = self.do_compile()
+            return self._compile_result
+
+    def do_compile(self) -> tuple[bool, str|None]:
```
Tagl, line 64:
```suggestion
    def do_compile(self) -> tuple[bool, Optional[str]]:
```

problemtools/run/source.py
```diff
@@ -80,24 +81,18 @@ def __init__(self, path, language, work_dir=None, include_dir=None):
...
 
-    def code_size(self):
+    def code_size(self) -> int:
         return sum(os.path.getsize(x) for x in self.src)
 
 
-    _compile_result = None
-
-    def compile(self):
+    def do_compile(self) -> tuple[bool, str|None]:
```
Tagl, line 88:
```suggestion
    def do_compile(self) -> tuple[bool, Optional[str]]:
```

problemtools/run/viva.py
```diff
@@ -22,26 +22,22 @@ def __init__(self, path):
...
 
 
     def __str__(self):
         """String representation"""
         return '%s' % (self.args[0])
 
 
-    _compile_result = None
-    def compile(self):
+    def do_compile(self) -> tuple[bool, str|None]:
```
Tagl, line 33:
```suggestion
    def do_compile(self) -> tuple[bool, Optional[str]]:
```

problemtools/verifyproblem.py
```diff
@@ -87,6 +91,15 @@ class VerifyError(Exception):
     pass
 
 
+class Context:
+    def __init__(self, args: argparse.Namespace, executor: ThreadPoolExecutor|None) -> None:
```
Tagl, line 95:
```suggestion
    def __init__(self, args: argparse.Namespace, executor: Optional[ThreadPoolExecutor]) -> None:
```

problemtools/verifyproblem.py
```diff
@@ -54,7 +58,7 @@ def __init__(self, verdict: str, score: float|None=None, reason: str|None=None,
         self.runtime_testcase: TestCase|None = None
         self.runtime = -1.0
         self.ac_runtime = -1.0
-        self.ac_runtime_testcase = None
+        self.ac_runtime_testcase: TestCase|None = None
```
Tagl, line 61:
```suggestion
        self.ac_runtime_testcase: Optional[TestCase] = None
```

problemtools/verifyproblem.py
```diff
@@ -87,6 +91,15 @@ class VerifyError(Exception):
     pass
 
 
+class Context:
+    def __init__(self, args: argparse.Namespace, executor: ThreadPoolExecutor|None) -> None:
+        self.data_filter: Pattern[str] = args.data_filter
+        self.submission_filter: Pattern[str] = args.submission_filter
+        self.fixed_timelim: int|None = args.fixed_timelim
```
Tagl, line 98:
```suggestion
        self.fixed_timelim: Optional[int] = args.fixed_timelim
```

problemtools/verifyproblem.py
```diff
@@ -422,7 +424,7 @@ def get_subgroups(self) -> list[TestCaseGroup]:
         return [child for child in self._items if isinstance(child, TestCaseGroup)]
 
 
-    def get_subgroup(self, name):
+    def get_subgroup(self, name: str) -> TestCaseGroup|None:
```
Tagl, line 427:
```suggestion
    def get_subgroup(self, name: str) -> Optional[TestCaseGroup]:
```

problemtools/verifyproblem.py
```diff
@@ -706,12 +711,12 @@ def __init__(self, problem: Problem):
     def __str__(self) -> str:
         return 'problem configuration'
 
-    def get(self, key: str|None=None):
+    def get(self, key: str|None=None) -> Any:
```
Tagl, line 714:
```suggestion
    def get(self, key: Optional[str]=None) -> Any:
```

problemtools/verifyproblem.py
```diff
@@ -1643,6 +1670,114 @@ def validate(self, testcase: TestCase, submission_output: str) -> SubmissionResu
...
+                return [item]
+        elif item not in self._done_groups:
+            ret = []
+            for child in item.get_testcases() + item.get_subgroups():
+                ret.extend(self._gather_testcases(child))
+            return ret
+        else:
+            return []
+
+    def _next_job(self) -> TestCase|None:
```
Tagl, line 1759:
```suggestion
    def _next_job(self) -> Optional[TestCase]:
```

problemtools/run/__init__.py
simonlindholm: I think `|None` is the more modern version (added in 3.10)? It matches what the rest of the code base uses, too. (I do agree personally that `|None` is a bit ugly.)

simonlindholm:
> Good suggestion! Why the letter j? Just because t is in use?

-j is the customary flag for threading. Also used by e.g. `make`, and by `generatedata.py` in this repo.

problemtools/run/__init__.py
Tagl: I was unaware of that and have now read through the following were this is debated: https://discuss.python.org/t/clarification-for-pep-604-is-foo-int-none-to-replace-all-use-of-foo-optional-int/26945/44

You can ignore all my suggestions then, I have no strong preference on this at all.

pehrsoderman (approved):

PR #268, opened 2024-07-24
Change team to user for default validator

niemela:
The problem package format is not only relevant for contests, but can also be used for e.g. education.

Changing "team" to the more generic "user"

### DIFF
--- a/support/default_validator/default_validator.cc
+++ b/support/default_validator/default_validator.cc
@@ -65,7 +65,7 @@ FILE *openfeedback(const char *feedbackdir, const char *feedback, const char *wh
 	return res;
 }
 
-const char *USAGE = "Usage: %s judge_in judge_ans feedback_file [options] < team_out";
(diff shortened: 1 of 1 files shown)


Tagl:
Do you want to update the example problem guess so that its validator uses user instead of author? Or is author fine?

niemela:
> Do you want to update the example problem guess so that its validator uses user instead of author? Or is author fine?

Yes, that's a good idea.

(IMO "author" is one of those terminology choices I *really* regret 15-20 years later 😄)

niemela:
> > Do you want to update the example problem guess so that its validator uses user instead of author? Or is author fine?
> 
> Yes, that's a good idea.

But, author is only used in the code, not in any output.... I think this is fine as is.

Tagl (approved):

pehrsoderman (approved):

PR #273, opened 2024-08-17
GCC should use gnu17

pehrsoderman:


mpsijm:
Thanks for your recent updates to `languages.yaml`! :smile: Just to double-check, https://open.kattis.com/languages/c still says `-std=gnu23`, is that intentional?

(This may be a similar thing as what I mentioned in https://github.com/Kattis/problemtools/pull/257#issue-2148575533, where the Kotlin version at https://open.kattis.com/languages/kotlin is still at 1.8, even though Kattis internals use(d?) 1.3)

PR #274, opened 2024-08-17
Markdown support

Matistjati:
supported for both pdf and html:
- problem name and id are automatically inserted at the top of the statement. Spec that we automatically prepend it from problem.yaml
- samples
- interactive samples
- {{nextsample}} and {{remainingsamples}}
- commonmark-style images
- "common" footnotes. extended commonmark syntax
- "common" tables. extended commonmark syntax
- inline and display math with `$a+b$` and `$$a+b$$`

A PR of the spec can be found here https://github.com/Kattis/problem-package-format/pull/386.

todo in follow-up pr:
- svgs, both for md and html. need to be sanitized
- better CSS class names for Kattis 
- CSS styling closer to Kattis

not planned to be added in this PR (maybe ever):
- tex -> html using pandoc (remove plastex)
- tex -> pdf using pandoc (we will keep using pdflatex for this)
- pdf images (they look terrible)
- webp images (poor support)


Matistjati:
@pehrsoderman want to take a look? I think this PR is ready for a review. Here are the statements i tested it on:

https://github.com/Matistjati/kattis-markdown-examples

In general, most statements don't look perfect right off the bat, mostly because they come from their own markdown dialect. However, I believe that all ugliness can be fixed by small tweaks.

problemtools/statement_common.py
```diff
@@ -0,0 +1,279 @@
...
+
+def get_problem_name(problem: str, language: Optional[str]) -> Optional[str]:
+    """Load problem.yaml to get problem name"""
+    if language is None:
+        language = "en"
+    with verifyproblem.Problem(problem) as prob:
+        config = verifyproblem.ProblemConfig(prob)
+    if not config.check(None):
+        raise Exception("Invalid problem.yaml")
+    names = config.get("name")
```
gkreitz, line 99: This will break after #286, but you're probably aware of that. :)

problemtools/statement_common.py
```diff
@@ -0,0 +1,279 @@
...
+        </table>""" % ({"case": casenum, "input": html.escape(sample_input),
+                        "output": html.escape(sample_output)})
+
+    if to_pdf:
+        # If pdf, convert to markdown
+        with tempfile.NamedTemporaryFile(mode='w', suffix=".html") as temp_file:
+            temp_file.write(sample)
+            temp_file.flush()
+            command = ["pandoc", temp_file.name, "-t" , "markdown"]
+            return subprocess.run(command, capture_output=True, text=True,
```
gkreitz, line 213: A bit of a corner case, but what happens if the sample contains latex (or, more likely,  latex control characters like `$`)? I'm assuming pandoc won't escape that when converting HTML to markdown, so won't that then later mess up the PDF rendering once you've injected this markdown into the document and are trying to `pandoc` into a pdf?

problemtools/statement_common.py
```diff
@@ -0,0 +1,279 @@
...
+                print(f"Warning: Interaction had unknown prefix {interaction[0]}")
+            lines.append(r"""
+                            \begin{table}[H]
+                            %(justify)s\begin{tabular}{|p{0.6\textwidth}|}
+                                \hline
+                                %(text)s \\
+                                \hline
+                            \end{tabular}
+                        \end{table}""" % {"justify": "" if left else "\\hspace*{\\fill}\n",
+                                            "text": data})
```
gkreitz, line 265: `data` probably needs to be escaped (for both markdown *and* latex, I guess)?

problemtools/statement_common.py
```diff
@@ -0,0 +1,279 @@
...
+        else:
+            line_type = ""
+            if interaction[0] == '>':
+                line_type = "sampleinteractionwrite"
+            elif interaction[0] == '<':
+                line_type = "sampleinteractionread"
+            else:
+                print(f"Warning: Interaction had unknown prefix {interaction[0]}")
+            lines.append(f"""<div class="{line_type}"><pre>{data}</pre></div>""")
```
gkreitz, line 274: `data` needs to be escaped (for html, so that's easy at least).

gkreitz (changes requested):
I took a pass over this. As mentioned on slack, it seems like pandoc makes no security guarantees on html output and recommends using a sanitizer (https://pandoc.org/chunkedhtml-demo/19-a-note-on-security.html). That's gonna get messy, I'm not a huge fan of any of the python options. If anyone happens to look at this PR and has any recommendations, feel free to chime in.

Looking through, I think I spotted a number of places where we need to be more careful with escaping stuff we inject (problem name, sample, and just as an extra precaution, problem id). I'm not quite sure how to do that cleanly, t.b.h. If I understand the code correctly, it seems like when pandoc goes from markdown to pdf, it goes via LaTeX, which means that inserting plain LaTeX in the markdown works. That dual-step translation makes me wonder how to sanely escape stuff like `\` or `$` when we inject it (or, as an extreme example, the sample of a problem about parsing latex :)).

problemtools/problem2pdf.py
```diff
@@ -5,16 +5,82 @@
...
+
+    statement_path = statement_common.find_statement(problem_root, extension="md", language=options.language)
+
+    if not os.path.isfile(statement_path):
+        raise Exception(f"Error! {statement_path} is not a file")
+
+    statement_common.assert_images_are_valid_md(statement_path)
+
+    templatepaths = [os.path.join(os.path.dirname(__file__), 'templates/markdown_pdf'),
+                    os.path.join(os.path.dirname(__file__), '../templates/markdown_pdf'),
```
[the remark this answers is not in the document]
Matistjati: It was in fact not needed.

problemtools/statement_common.py
Matistjati: Yup, talked with Rasmus and Hugo to hopefully work something out

problemtools/md2html.py
```diff
@@ -0,0 +1,122 @@
...
+           title=problem_name or "Missing problem name",
+           problemid=problembase)
+
+    samples = statement_common.format_samples(problem, to_pdf=False)
+
+    # Insert samples at {{nextsample}} and {{remainingsamples}}
+    html_template, remaining_samples = statement_common.inject_samples(html_template, samples, "")
+
+    # Insert the remaining samples at the bottom
+    if FOOTNOTES_STRING in html_template:
```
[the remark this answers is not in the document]
Matistjati: Done

Matistjati:
Known issues:
- (PDF) Samples might overflow
image
- (PDF) If you use all but the simplest possible table, the outline will not work properly. I tried a workaround, which works for some tables, but we will need a LaTeX wiz (not me) to account for all tables using macros. See https://github.com/jgm/pandoc/issues/922
image
- (HTML) If you place `{{remainingsamples}}` in the text, you end up with something like
image
- SVGs are currently not supported. Will be fixed in a follow-up PR.
- Does not support multipass samples

problemtools/statement_common.py
Matistjati: Done

problemtools/statement_common.py
Matistjati: Done

problemtools/statement_common.py
Matistjati: Painstakingly, this has been fixed.

problemtools/statement_common.py
```diff
@@ -0,0 +1,396 @@
...
+        casenum += 1
+
+    return samples
+
+def escape_latex_char(char: str) -> str:
+    if len(char) != 1:
+        raise ValueError("Input must be a single character.")
+
+    replacements = {
+        "\\": "\\textbackslash{}",
```
gkreitz, line 204: Nitpick/feature suggestion: Python has "raw string literals" for exactly this use case. If you write these strings as `r'\textbackslash{}'` you avoid double escaping almost all the backslashes here.

problemtools/statement_common.py
```diff
@@ -0,0 +1,396 @@
...
+                    table += list(row)
+                    table += list(row)
+
+                # Add stuff to write to this line while it fits
+                curr_vspace = 0
+                curr_line = ""
+                # Must fit in both Markdown table and LaTeX table
+                while len(line) and \
+                    len(curr_line)+1<chars_per_row and \
+                    curr_vspace + ascii_char_widths[line[0]] < space_per_row:
```
gkreitz, line 313: It looks to me like this would crash for characters not in `ascii_char_widths`. I don't think it's feasible to build a table of all possible characters. IIRC from typography, I think the best you can do for a character not in the table is to look up with width of the letter m.

problemtools/statement_common.py
```diff
@@ -0,0 +1,396 @@
...
+    ascii_char_widths = {' ': 3.33333, '!': 3.2, '"': 6.2, '#': 9.6, '$': 5.9, '%': 9.6, '&': 9.0, "'": 3.2, '(': 4.5, ')': 4.5, '*': 5.8, '+': 9.0, ',': 6.2, '-': 6.5, '.': 5, '/': 5.8, '0': 5.8, '1': 5.8, '2': 5.8, '3': 5.8, '4': 5.8, '5': 5.8, '6': 5.8, '7': 5.8, '8': 5.8, '9': 5.8, ':': 3.2, ';': 3.2, '<': 8.9, '=': 8.9, '>': 8.9, '?': 5.4, '@': 8.9, 'A': 7.50002, 'B': 7.08336, 'C': 7.22223, 'D': 7.6389, 'E': 6.80557, 'F': 6.5278, 'G': 7.84723, 'H': 7.50002, 'I': 3.61111, 'J': 5.1389, 'K': 8.5, 'L': 6.25002, 'M': 9.16669, 'N': 7.50002, 'O': 8.5, 'P': 6.80557, 'Q': 8.5, 'R': 7.36111, 'S': 5.55557, 'T': 7.22223, 'U': 7.50002, 'V': 7.50002, 'W': 10.2778, 'X': 7.50002, 'Y': 7.50002, 'Z': 6.11111, '[': 6.2, '\\': 6.0, ']': 6.2, '^': 6.5, '_': 8.6, '`': 5.8, 'a': 5.8, 'b': 5.55557, 'c': 4.44444, 'd': 5.55557, 'e': 4.44444, 'f': 3.05557, 'g': 5.8, 'h': 5.55557, 'i': 3.2, 'j': 3.05557, 'k': 5.2778, 'l': 3.2, 'm': 9.6, 'n': 5.55557, 'o': 5.8, 'p': 5.55557, 'q': 5.27779, 'r': 3.91667, 's': 3.94444, 't': 4.5, 'u': 5.55557, 'v': 5.2778, 'w': 7.22223, 'x': 5.2778, 'y': 5.2778, 'z': 4.44444, '{': 5.8, '|': 3.3, '}': 5.8, '~': 6.5}
```
gkreitz, line 286: Ouch, what a table to build. Don't we want to use a monospace font for this, though?

problemtools/statement_common.py
```diff
@@ -0,0 +1,396 @@
...
+            elif interaction[0] == '<':
+                left = False
+            else:
+                left = True
+                print(f"Warning: Interaction had unknown prefix {interaction[0]}")
+            lines.append(r"""
+                            \begin{table}[H]
+                            %(justify)s\begin{tabular}{|p{0.6\textwidth}|}
+                                \hline
+                                %(text)s \\
```
gkreitz, line 378: This looks like it will still break if there are latex control characters in the sample?

gkreitz (commented):
I took another pass over it, as requested. This is getting close to ready for merging. The main problem I see are a couple of places where we need to escape latex characters (the rendering path of markdown -> latex -> pdf that pandoc does is funky...).

problemtools/statement_common.py
Matistjati: Fixed by depending on normal tex -> pdf in problemtools

problemtools/statement_common.py
Matistjati: Fixed by depending on normal tex -> pdf in problemtools

problemtools/statement_common.py
Matistjati: Fixed by depending on normal tex -> pdf in problemtools

gkreitz (approved):
Looks good, code just had two small things. Tests failing due to UUID should be fixed by #307 merging (and then rebasing on that).

For example problems, I'm ok with merging this as-is, but if so we should do a follow-up soon where we convert the problems to 2023-07.

tests/test_markdown.py
```diff
@@ -0,0 +1,47 @@
...
+from tests.test_xss import render, renderpdf
+from problemtools.statement_util import find_footnotes
+import pytest
+
+
+def test_sample_escaping():
+    problem_path = Path(__file__).parent / 'problems' / 'specialcharacterssample'
+    html = render(problem_path)
+    all_printable = r"""0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ!"#$%&'()*+,-./:;<=>?@[\]^_`{|}~"""
+    # nh3 will escape &, < and >
```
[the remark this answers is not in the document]
Matistjati: You are right, fixed

problemtools/statement_util.py
```diff
@@ -0,0 +1,263 @@
...
+    for interaction in sample_interaction:
+        data = html.escape(interaction[1:])
+        line_type = ''
+        if interaction[0] == '>':
+            line_type = 'sampleinteractionwrite'
+        elif interaction[0] == '<':
+            line_type = 'sampleinteractionread'
+        else:
+            print(f'Warning: Interaction had unknown prefix {interaction[0]}')
+        lines.append(f"""<div class="{line_type}"><pre>{html.escape(data)}</pre></div>""")
```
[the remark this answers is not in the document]
Matistjati: Yup, fixed

PR #276, opened 2024-08-19
Allow symlinks to test cases with other output validator flags and…

### LINKED ISSUE #244: Symlinks for reuse of input files with different output validator flags
https://github.com/Kattis/problemtools/blob/92dcbf3fab45ed4c13438511df98ec82fec512f9/problemtools/verifyproblem.py#L212C16-L212C16

Should the above not be a warning, rather than an error? I don't see why it would be an error.
Using symlinks is a nice way of reusing the input/answer files.
For example, you can use the same test cases in two different groups. In one of them validation is case insensitive and in the other a stricter case sensitive validation is applied.

Tagl:
..warn if output validators fails to compile. Closes #244 

Note this may be a breaking change for some problem based on discussion in issue.
However that problem is not being validated "correctly" since the output validator generally should behave differently with different arguments passed to it, rendering reuse impossible.

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -183,8 +183,8 @@ class TestCase(ProblemAspect):
         self._problem = problem
         self.testcasegroup = testcasegroup
         self.reuse_result_from: TestCase|None = None
-        self.counter = len(problem.testcase_by_infile)
-        problem.testcase_by_infile[self.infile] = self
+        self.counter = len(problem.testcase_by_input)
+        problem.testcase_by_input[(self.infile, self.testcasegroup.config['output_validator_flags'])] = self
 
     def check_newlines(self, filename: str) -> None:
         with open(filename, 'rb') as f:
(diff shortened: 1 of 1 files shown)


problemtools/verifyproblem.py
```diff
@@ -264,12 +264,9 @@ def _check_symlinks(self) -> bool:
         if ans_target != f'{in_target[:-3]}.ans':
             self.error(f"Symbolic link '{nicepath}' must have a corresponding link for answer file")
             return False
-        if self.reuse_result_from is None:
+        if not nicepath.startswith('data'):
```
simonlindholm, line 267: Isn't the previous check more accurate?

problemtools/verifyproblem.py
```diff
@@ -1687,6 +1686,8 @@ def validate(self, testcase: TestCase, submission_output: str) -> SubmissionResu
                 shutil.rmtree(validator_output)
                 if res.verdict != 'AC':
                     return res
+            else:
+                self.warning(f"Compilation failed for {val.name}: {compile_res[1]}")
```
simonlindholm, line 1690: Won't this result in a lot of warning spam (one log line per submission per test case)? IIRC we already give an error in an earlier phase.

Tagl: I had an error and got no information before adding this. It was very confusing.
I will double check though.

problemtools/verifyproblem.py
Tagl: How so?

simonlindholm: Sorry for late reply, but:
- this covers e.g. `data2` too
- `nicepath` is the path prior to symlink resolution

problemtools/verifyproblem.py
gkreitz: I've fixed this in a less spammy way in #319

gkreitz:
I've reworked the handling of testcase reuse completely in #411. Symlinks now have no special treatment - we determine when we can safely reuse results and do so when test cases are identical. Symlinks are now allow between non-identical test cases (e.g., to save space). Closing this PR, as I feel like the approach in #411 is a better long-term fix.

PR #283, opened 2025-03-02
Update default_validator; add custom error when user output is empty

JoelNiemela:
When the user output file is completely empty, add the error message "user output was empty"; when the user output file contains only whitespace, add the error message "user output contained no tokens".

Other changes:
 - Rewrite `if (...) stmt;` as `if (...) { stmt; }`
 - Explicitly assign zero to `judgeans_pos` and `stdin_pos` (line 17). Previously, they were implicitly initialized to zero (since they are global variables).
 - Initialize `judgeans_line` and `stdin_line` to `1` (line 18). Previously, they were set to `1` in `main()`, before the `while (true)` loop.
 - Add some whitespace to be more consistent.

### DIFF
--- a/support/default_validator/default_validator.cc
+++ b/support/default_validator/default_validator.cc
@@ -14,14 +14,13 @@ const int EXIT_WA = 43;
 std::ifstream judgein, judgeans;
 FILE *judgemessage = NULL;
 FILE *diffpos = NULL;
-int judgeans_pos, stdin_pos;
-int judgeans_line, stdin_line;
+int judgeans_pos = 0, stdin_pos = 0;
+int judgeans_line = 1, stdin_line = 1;
 
 void wrong_answer(const char *err, ...) {
 	va_list pvar;
 	va_start(pvar, err);
-	fprintf(judgemessage, "Wrong answer on line %d of output (corresponding to line %d in answer file)\n",
-			stdin_line, judgeans_line);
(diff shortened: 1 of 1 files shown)


support/default_validator/default_validator.cc
```diff
@@ -134,20 +133,35 @@ int main(int argc, char **argv) {
...
-			wrong_answer("User EOF while judge had more output\n(Next judge token: %s)", judge.c_str());
+			if (token == 0) {
+				if (stdin_pos == 0) {
+					wrong_answer(
+						"User EOF while judge had more output; user output was empty.\n(Next judge token: %s)",
+						judge.c_str()
+					);
+				} else {
+					wrong_answer(
+						"User EOF while judge had more output; user output contained no tokens.\n(Next judge token: %s)",
```
gkreitz, line 149: This happens if and only if the user output is only whitespace, right? To me "user output contained only whitespace" would be more clear than "contained no tokens".

gkreitz (approved):
Thanks, looks good!

I suggested one minor change to a message. Happy to merge this as-is too (if you prefer the current message, or if I my suggested phrasing is incorrect).

support/default_validator/default_validator.cc
JoelNiemela: Yes. That feels cleared to me too, fixed.

PR #286, opened 2025-03-06
Implement new system for defining problem-formats

square-cylinder:
The old system for working with the different aspects of verifying a problem were not flexible enough to use for multiple formats, so a new system has been developed to make this more flexible.

# Overview
- All checkable aspects of a problem now have to subclass ProblemPart
- All current classes that were used to verify parts of problems now inherit ProblemPart (except for TestCaseGroup, where a new class ProblemTestCases was added for verification)
- Some minor changes were done in most of these classes to make them work better with the new system, mostly changing how they access data between each other
- When instantiating a Problem, will now have to provide a "part_mapping" which is a dictionary from string (categories) to classes which are checked for that category
- Added a commandline argument to ensure that problems follow a certain format (or 'automatic' if it should be detected from metadata)
- Added handling of KeyboardInterrupt
- Removed pre-legacy version of \problemname{Name}, %% plainproblemname:Name

# Description of new system
## Definition of a problemformat
The system for describing a problem-format works by providing a dictionary of categories to classes (types), similar to how "part_mapping" worked previously. For example for the legacy configuration right now we have:
```py
{
        'config':       [ProblemConfig],
        'statement':    [ProblemStatementLegacy, Attachments],
        'validators':   [InputValidators, OutputValidators],
        'graders':      [Graders],
        'data':         [ProblemTestCases],
        'submissions':  [Submissions],
}
```
This means we have the validation steps 'config', 'statement', 'validators', 'graders', 'data', and 'submissions'. These can be included via the commandline-option -p like before. The list of classes are which classes will be checked for each of these parts.

## class ProblemPart
This is possible to do via new data that is defined in ProblemPart for each part that can be included. Each problem-part will have to override a class-variable `PART_NAME`  which is a string representing the name of the part. There is a new system for communicating data between different problem-parts through the problem.

The problem has a dictionary for each problem-part where properties can be set by each ProblemPart subclass. This data can then be accessed through for example `self.problem.get('statement', 'config')`. To set a property for the problem-part you can write for example `self.set_prop('foo', 'bar')`, which would set the property 'foo' to 'bar'.

It is also possible to access `problem.classes['statement']` but the intention is to try to avoid this whenever possible, and rely only on the problem-part-properties instead. Hopefully we will be able to find a way to convert the current classes to follow this in future PR's. The thought-process is that these properties will be the only communication-channel between different parts of the problem to simplify the interactions.

Since some parts require information from other parts in order to be initialized, a system for keeping track of initialization-dependencies was implemented. In each ProblemPart subclass, the method `depends_on() -> set[type]` is possible to be overridden. For every class that is returned from this method, the problem-class will make sure that a subclass of that class is initialized before this one. For example in the old code, ProblemConfig required some information from ProblemStatement in order to be initialized so ProblemConfig will define:
```py
def depends_on() -> set[type]:
    return {ProblemStatement}
```
The Problem-class will then ensure that when setting up all the classes, it will resolve all dependencies to the classes before initializing them.

## class Problem
Surprisingly little was changed for the Problem-class. The general logic was abstracted a bit to allow for multiple problem-formats, but the overall structure is similar to how it was previously. The Problem class will now require a problem-format in its constructor.

## ProblemTestCases
The TestCaseGroup class was difficult to use with the new system since it recursively contains copies of itself, so a new class ProblemTestCases was added as a kind of wrapper around the TestCaseGroup class. It provides a decent example of a ProblemPart-subclass can look since it contains a dependency, is pretty small and uses the new communication-system with the Problem class.
```py
class ProblemTestCases(ProblemPart):
    
    PART_NAME = 'testdata'
    
    @staticmethod
    def depends_on():
        return {ProblemConfig}

    def setup(self):
        self.set_prop('testcase_by_infile', {})
        self.set_prop('root_group', TestCaseGroup(self.problem, self.PART_NAME))
        self.set_prop('is_interactive', 'interactive' in self.problem.get(ProblemConfig, 'data')['validation-params'])
        self.set_prop('is_scoring', self.problem.get(ProblemConfig, 'data')['type'] == 'scoring')

    def check(self, context: Context) -> bool:
        return self.problem.get(ProblemTestCases, 'root_group').check(context)
```

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -184,6 +184,7 @@ class ProblemPart(ProblemAspect):
     Note that this will only ensure that the specified classes are initialized before this one, but
     they might be checked in a different order.
     """
+    @staticmethod
     def depends_on() -> set[type]:
         return set()
     
@@ -784,6 +785,8 @@ class ProblemStatement2023_07(ProblemStatement):
 
 class ProblemConfig(ProblemPart):
     PART_NAME = 'config'
+
+    @staticmethod
     def depends_on():
         return {ProblemStatement}
 
@@ -943,6 +946,7 @@ class ProblemTestCases(ProblemPart):
     
     PART_NAME = 'testdata'
     
+    @staticmethod
     def depends_on():
         return {ProblemConfig}
 
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -525,7 +525,7 @@ class TestCaseGroup(ProblemAspect):
                     self.error(f"Invalid score range '{score_range}': minimum score cannot be greater than maximum score")
             except VerifyError:
                 raise
-            except:
+            except Exception:
                 self.error(f"Invalid format '{score_range}' for range: must be exactly two floats")
 
         if self._parent is None:
@@ -697,14 +697,14 @@ class ProblemStatement(ProblemPart):
 
     def setup(self):
         self.debug('  Loading problem statement')
-        self.languages = []
-        glob_path = os.path.join(self.problem.probdir, 'problem_statement', 'problem.')
-        if glob.glob(glob_path + 'tex'):
-            self.languages.append('')
-        for f in glob.glob(glob_path + '[a-z][a-z].tex'):
-            m = re.search("problem.([a-z][a-z]).tex$", f)
-            assert m
-            self.languages.append(m.group(1))
+        self.languages = self.list_languages()
+        #glob_path = os.path.join(self.problem.probdir, 'problem_statement', 'problem.')
+        #if glob.glob(glob_path + 'tex'):
+        #    self.languages.append('')
+        #for f in glob.glob(glob_path + '[a-z][a-z].tex'):
+        #    m = re.search("problem.([a-z][a-z]).tex$", f)
+        #    assert m
+        #    self.languages.append(m.group(1))
 
         self.set_prop('config', self.get_config())
 
@@ -715,8 +715,9 @@ class ProblemStatement(ProblemPart):
 
         if not self.languages:
             self.error('No problem statements found (expected problem.tex or problem.[a-z][a-z].tex in problem_statement directory)')
-        if '' in self.languages and 'en' in self.languages:
-            self.error("Can't supply both problem.tex and problem.en.tex")
+        for lang, count in collections.Counter(self.languages).items():
+            if count > 1:
+                self.error(f"Can't supply multiple statements of the same language ({lang}).")
 
         for lang in self.languages:
             try:
@@ -764,6 +765,35 @@ class ProblemStatement(ProblemPart):
                     ret[dest][lang] = hit.group(1).strip()
         return ret
 
+    def list_languages(self) -> list[str]:
+        print(type(self))
+        raise NotImplementedError('Need to specialize ProblemStatement to list languages')
+
+class ProblemStatementLegacy(ProblemStatement):
+    def list_languages(self) -> list[str]:
+        languages = []
+        glob_path = os.path.join(self.problem.probdir, 'problem_statement', 'problem.')
+        if glob.glob(glob_path + 'tex'):
+            languages.append('en')
+        for f in glob.glob(glob_path + '[a-z][a-z].tex'):
+            m = re.search("problem.([a-z][a-z]).tex$", f)
+            assert m
+            languages.append(m.group(1))
+        return languages
+
+class ProblemStatement2023_07(ProblemStatement):
+    def list_languages(self) -> list[str]:
+        languages = []
+        for ext in ('tex', 'md'):
+            glob_path = os.path.join(self.problem.probdir, 'problem_statement', 'problem.')
+            if glob.glob(glob_path + ext):
+                languages.append('en')
+            for f in glob.glob(glob_path + '[a-z][a-z].' + ext):
+                m = re.search(f"problem.([a-z][a-z]).{ext}$", f)
+                assert m
+                languages.append(m.group(1))
+        return languages
+
 class ProblemConfig(ProblemPart):
     PART_NAME = 'config'
     DEPENDS_ON = {ProblemStatement}
@@ -788,7 +818,7 @@ class ProblemConfig(ProblemPart):
                 self.error(str(e))
 
         # Add config items from problem statement e.g. name
-        self._data.update(self.problem.classes['statement'].get_config())
+        self._data.update(self.problem.get(ProblemStatement.PART_NAME, 'config'))
 
         # Populate rights_owner unless license is public domain
         if 'rights_owner' not in self._data and self._data.get('license') != 'public domain':
@@ -1717,6 +1747,7 @@ class Problem(ProblemAspect):
         super().__init__(self.shortname)
         self.language_config = languages.load_language_config()
         self.data = {}
+        self.debug(f'Problem-format: {parts}')
 
     def get(self, part, key, default=None) -> Any:
         if part not in self.data.keys():
@@ -1737,10 +1768,15 @@ class Problem(ProblemAspect):
         def init(_class):
             if _class.PART_NAME in initialized:
                 return
+            # A bit ugly but want to allow for subclasses
             for dependency in _class.DEPENDS_ON:
-                if dependency not in self.aspects:
+                for cl in self.aspects:
+                    if issubclass(cl, dependency):
+                        init(cl)
+                        break
+                else:
                     raise NotImplementedError(f'Part "{_class.PART_NAME}" depends on part "{dependency.PART_NAME}" which is not included in problem-format')
-                init(dependency)
+            self.debug(f'Initializing {_class.PART_NAME} ({_class})')
             self.classes[_class.PART_NAME] = _class(self)
             initialized.add(_class.PART_NAME)
 
@@ -1837,14 +1873,14 @@ def part_argument(s: str) -> str:
 PROBLEM_FORMATS = {
     'legacy': {
         'config':       [ProblemConfig],
-        'statement':    [ProblemStatement, Attachments],
+        'statement':    [ProblemStatementLegacy, Attachments],
         'validators':   [InputValidators, OutputValidators],
         'graders':      [Graders],
         'data':         [ProblemTestCases],
         'submissions':  [Submissions],
     },
-    '2023-07': { # TODO
-
+    '2023-07': { # TODO: Add all the parts
+        'statement':    [ProblemStatement2023_07, Attachments],
     }
 }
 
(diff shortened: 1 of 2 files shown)


problemtools/verifyproblem.py
```diff
@@ -669,17 +711,92 @@ def all_datasets(self) -> list:
...
+
+        self.set_prop('config', self.get_config())
+
+    def check(self, context: Context) -> bool:
+        if self._check_res is not None:
+            return self._check_res
+        self._check_res = True
+
+        if not self.statements:
+            self.error(f'No problem statements found (expected file of one of following forms in folder problem_statement/: {', '.join(f'problem.{ext}, problem.[a-z][a-z].{ext}' for ext in self.EXTENSIONS)}')
```
gkreitz, line 735: This is a syntax error in earlier python versions (which is why the CI tests fail), so you'll want to rewrite this.

problemtools/verifyproblem.py
```diff
@@ -785,7 +904,7 @@ def check(self, context: Context) -> bool:
             self.error(f"Invalid value for grading.show_test_data_groups: {self._data['grading']['show_test_data_groups']}")
         elif self._data['grading']['show_test_data_groups'] and self._data['type'] == 'pass-fail':
             self.error("Showing test data groups is only supported for scoring problems, this is a pass-fail problem")
-        if self._data['type'] != 'pass-fail' and self._problem.testdata.has_custom_groups() and 'show_test_data_groups' not in self._origdata.get('grading', {}):
+        if self._data['type'] != 'pass-fail' and self.problem.get(ProblemTestCases, 'root_group').has_custom_groups() and 'show_test_data_groups' not in self._origdata.get('grading', {}):
```
gkreitz, line 907: Don't you need to depend on `ProblemTestCases` for this to work (reliably)? That also feels like it's a bit of a problem, as `ProblemTestCases` depends on this class.

problemtools/verifyproblem.py
```diff
@@ -1004,7 +1071,7 @@ def collect_flags(group: TestCaseGroup, flags: set[str]) -> None:
                     flags.add(group.config['input_validator_flags'])
                 for subgroup in group.get_subgroups():
                     collect_flags(subgroup, flags)
-            collect_flags(self._problem.testdata, all_flags)
+            collect_flags(self.problem.get(ProblemTestCases, 'root_group'), all_flags)
```
gkreitz, line 1074: Looks like this class needs to depend on `ProblemTestCases`

problemtools/verifyproblem.py
```diff
@@ -1202,12 +1268,12 @@ def check(self, context: Context) -> bool:
             if isinstance(v, run.SourceCode) and v.language.lang_id not in recommended_output_validator_languages:
                 self.warning('output validator language %s is not recommended' % v.language.name)
 
-        if self._problem.config.get('validation') == 'default' and self._validators:
+        if self.problem.classes['config'].get('validation') == 'default' and self._validators:
```
gkreitz, line 1271: These calls (plenty here, and in more places in this class) to `self.problem.classes['config'].get` should probably all be `self.problem.get(ProblemConfig, 'data').get`, right? And you also need to depend on `ProblemConfig` (I assume you need this with the current version too, to ensure that the `'config'` key is populated in `problem.classes`.)

problemtools/verifyproblem.py
```diff
@@ -1229,7 +1295,7 @@ def check(self, context: Context) -> bool:
                 f.write(case)
                 f.close()
                 rejected = False
-                for testcase in self._problem.testdata.get_all_testcases():
+                for testcase in self.problem.get(ProblemTestCases, 'root_group').get_all_testcases():
```
gkreitz, line 1298: This also needs to be added as a `depends_on`

problemtools/verifyproblem.py
```diff
@@ -1549,8 +1615,8 @@ def check_submission(self, sub, context: Context, expected_verdict: Verdict, tim
         else:
             timelim_low = timelim
 
-        with Runner(self._problem, sub, context, timelim, timelim_low, timelim_high) as runner:
-            result, result_low, result_high = self._problem.testdata.run_submission(sub, runner, context)
+        with Runner(self.problem, sub, context, timelim, timelim_low, timelim_high) as runner:
+            result, result_low, result_high = self.problem.get(ProblemTestCases, 'root_group').run_submission(sub, runner, context)
```
gkreitz, line 1619: Looks like there are missing `depends_on` here too (`ProblemTestCases` , `OutputValidators`, `ProblemConfig` if I read the code in this class correctly).

problemtools/verifyproblem.py
```diff
@@ -1660,14 +1726,31 @@ def check(self, context: Context) -> bool:
...
         self.shortname: str|None = os.path.basename(self.probdir)
         super().__init__(self.shortname)
         self.language_config = languages.load_language_config()
+        self.data = {}
+        self.debug(f'Problem-format: {parts}')
+
+    def get(self, part, key, default=None) -> Any:
+        if isinstance(part, type) and issubclass(part, ProblemPart):
+            part = part.PART_NAME
+        if part not in self.data:
```
gkreitz, line 1751: This feels like a footgun to me. Do you ever use `default`, I don't think I spotted any calls with it? It feels extremely error prone that calling `get` silently returns `None` if `part not in self.data` (i.e., that part was either a typo, or wasn't initialized yet).

I would prefer if trying to access data from an uninitialized part is always an error (in problemtools, not the problem we're testing :)). If you *really* need this functionality, I think it makes sense to force the caller to explicitly toggle it on (e.g,. `get(self, part, key, default=None, crash_if_part_uninitialized=True)`)

problemtools/verifyproblem.py
```diff
@@ -1676,19 +1759,28 @@ def __enter__(self) -> Problem:
...
+
+        def init(_class):
+            if _class.PART_NAME in initialized:
+                return
+            # A bit ugly but want to allow for subclasses
+            for dependency in _class.depends_on():
+                for cl in self.aspects:
+                    if issubclass(cl, dependency):
+                        init(cl)
+                        break
```
gkreitz, line 1774: This feels a bit scary. I would love if you either initialize *all* classes which are subclasses of `dependency`, or add a check ensuring that there is only a unique class in `self.aspects` which is a subclass of `dependency`.

problemtools/verifyproblem.py
```diff
@@ -1676,19 +1759,28 @@ def __enter__(self) -> Problem:
...
+                else:
+                    raise NotImplementedError(f'Part "{_class.PART_NAME}" depends on part "{dependency.PART_NAME}" which is not included in problem-format')
+            self.debug(f'Initializing {_class.PART_NAME} ({_class})')
+            self.classes[_class.PART_NAME] = _class(self)
```
gkreitz, line 1778: It would make sense to me if you did `assert _class.PART_NAME not in self.data` and then `self.data[_class.PART_NAME] = {}` just before this line (following up on my earlier comment re ownership of `problem.data`).

gkreitz (changes requested):
Thanks for the PR. Overall, I like the direction you're heading in here. The PR description is excellent, and helped with navigating the code. Thanks for that!

I think a major issue is that you seem to have missed declaring `depends_on` for several methods where I think you *do* in fact depend on other components. I'm guessing this went unnoticed as `Problem.get` silently returns `None` when this happens (plus, I guess it's "lucky" that `config` sorts early :)). Unless I'm missing something, this will be a bit tricky to fix, as I think some of the dependencies are circular.

I also added a few more nit-picky comments (typically accessing via `.classes`, and/or using hard-coded strings instead of `Foo.PART_NAME`), which are not blockers for merging, but I wrote them as I was working my way through the code (and while I wasn't sure I understood the PR), and left them in.

square-cylinder:
# Changes made since last time
## Information channel between problem-parts
First of all, if an invalid problem-part is indexed, this will now give an assertion-error. Instead of having an extra layer of key indexing so doing Problem.get(<part>, <key>), the indexing is now instead Problem.get(<part>) and this will return a dictionary so keys can be handled more ergonomically. The `set_prop` function was removed, because it gave a hint that the data could be modified in for example the `check` function which was never the intention. Instead, the setup function will now return a dictionary which is the data that gets exposed from a given ProblemPart specialization. The setup of the data is now entirely handled by the Problem class.

## Renamed depends_on() to setup_dependencies()
I noticed you had the impression that there were dependencies between parts when the data-access happened through for example the `check` function. This is not an issue as the dependencies are only relevant for the setup part, because the data should not be changed after that. I realize that the function-name might give the wrong impression, so it was changed to better reflect the intention.

## Initializing only one subclass of dependency
The code was changed to make sure exactly one subclass was initialized when initializing dependency, and an additional assertion was added to make sure no two classes had the same PART_NAME when being initialized.

## Indexing with string
Fixed all the indexing to be done with the constants instead of with the strings (I think), and changed all the `classes[ProblemConfig.PART_NAME]` to use `Problem.get(ProblemConfig)` instead. To make sure no class was using the ProblemConfig.get method anymore this method was removed entirely.

## Nested f-string
The nested f-string was changed into to two statements which should hopefully make the tests pass now.

## Changed testcase_by_infile to be a class-parameter of ProblemTestCases
Because this is changed as a part of the testcases being ran, this should not be a part of the Problem.get data. Therefore I added is as an instance-variable to ProblemTestCases and accessed it through `problem.classes`.

problemtools/verifyproblem.py
```diff
@@ -1676,19 +1759,33 @@ def __enter__(self) -> Problem:
...
+            # A bit ugly but want to allow for subclasses
+            for dependency in _class.setup_dependencies():
+                cnt = 0
+                for cl in self.aspects:
+                    if issubclass(cl, dependency):
+                        init(cl)
+                        cnt += 1
+                        break
+                if cnt != 1:
+                    raise NotImplementedError(f'Part "{_class.PART_NAME}" depends on part "{dependency.PART_NAME}" which showed up {cnt} times in problem-format (should have showed up exactly once)')
```
gkreitz, line 1779: This check looks broken, as you didn't remove the `break` above (it will only detect 0 vs. 1, but won't notice if it appears multiple times).

As you added the `assert`  below, we will still blow up if something occurs multiple times, so this bug is mostly harmless.

gkreitz (approved):
Thanks for the quick fixes and explanation! 

I'm a bit unsure why the unit tests blow up in the same way as previously, they still seem to run the old github action. I fixed the tests in #288, can you try rebase:ing this PR?

I added two minor comments. Neither is a blocker.

Firstly, it seems like 50-50 whether you use .get(key) or `[key]` when accessing the dict returned by `problem.get(...)`. I'm not sure if there's some subtle pattern I'm not seeing. Both work just fine, but as a reader of the code, using two different styles gets a tad confusing.

Secondly, there was a minor bug in a check, but thankfully you added defense in depth there, making the bug harmless (it should still be fixed, but it's not a blocker).

I'm setting approve now as this looks good enough, but I would love to see unit tests working. Can you try to rebase (and also delete the buggy `break`), and then we can merge this.

PR #295, opened 2025-04-10
Updated parsing of problem.yaml by using checks as described in the specification

Zazmuz:
Added checks for legacy and 2023-07 for parsing problem.yaml, making sure its correct and defaulting values. 
Made sure to keep the code itself simple so that eventual changes and upkeep to it will be easy to go through with. 
It has been tested on most PO problems from the last 5 years as well as islandic problems. 

For the new format it has not been tested as much (due to lack of available configs), only a few edgecases and few self made configs.

If there is anything that looks weird or any questions I will be available.

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -577,7 +577,7 @@ class TestCaseGroup(ProblemAspect):
                 for filename in files:
                     filepath = os.path.join(root, filename)
                     if filepath.endswith('.in') and not os.path.islink(filepath):
-                        md5 = hashlib.md5()
+                        md5 = hashlib.md5(usedforsecurity=False)
                         with open(filepath, 'rb') as f:
                             for buf in iter(lambda: f.read(1024), b''):
                                 md5.update(buf)
@@ -961,6 +961,8 @@ class ProblemConfigLegacy(ProblemConfigBase):
             elif self._data['grading']['show_test_data_groups'] and self._data['type'] == 'pass-fail':
                 self.error("Showing test data groups is only supported for scoring problems, this is a pass-fail problem")
         else:
+            if self._data['type'] != 'pass-fail' and self.problem.get(ProblemTestCases)['root_group'].has_custom_groups() and 'show_test_data_groups' not in self._data.get('grading', {}):
+                self.warning("Problem has custom testcase groups, but does not specify a value for grading.show_test_data_groups; defaulting to false")
             self._data['grading']['show_test_data_groups'] = False
 
         if 'on_reject' in self._data['grading']:
@@ -1023,11 +1025,6 @@ class ProblemConfigLegacy(ProblemConfigBase):
             if 'source_url' in self._data:
                 self.error('source needs to be defined when source_url is defined')
 
-    def _check_custom_groups(self):
-        # TODO: implement the check
-        pass
-        #if self._data['type'] != 'pass-fail' and self.problem.get(ProblemTestCases)['root_group'].has_custom_groups() and 'show_test_data_groups' not in self._origdata.get('grading', {}):
-        #    self.warning("Problem has custom testcase groups, but does not specify a value for grading.show_test_data_groups; defaulting to false")
 
     def _check_unknown_fields(self):
         known = {'problem_format_version', 'type', 'name', 'uuid', 'author', 'source', 'source_url', 'license', 'rights_owner', 'limits', 'validation', 'validation-type', 'validation-params', 'validator_flags', 'grading', 'keywords'}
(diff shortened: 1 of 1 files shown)


problemtools/verifyproblem.py
```diff
@@ -769,73 +773,223 @@ def check(self, context: Context) -> bool:
...
-        # Add config items from problem statement e.g. name
-        self._data.update(self.problem.get(ProblemStatement))
+        self._data.setdefault('license', 'unknown')
+        if type(self._data['license']) is str:
+            self._data['license'] = self._data['license'].lower()
+        if self._data['license'] not in self._VALID_LICENSES:
+            self.error(f'License needs to be one of: {self._VALID_LICENSES}')
+        if self._data['license'] == 'unknown':
+            self.warning("License is 'unknown'")
+    def fix_config_structure(self):
```
gkreitz, line 832: It looks to me like `ProblemConfigBase` is an abstract base class, and that this is an abstract method? If so, consider inheriting from `ABC` and use the decorator `@abstractmethod` here.

problemtools/verifyproblem.py
```diff
@@ -769,73 +773,223 @@ def check(self, context: Context) -> bool:
...
-        # Populate rights_owner unless license is public domain
-        if 'rights_owner' not in self._data and self._data.get('license') != 'public domain':
-            if 'author' in self._data:
+    def check_uuid(self, uuid: str) -> None:
+        if not re.match(r'^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$', uuid):
+            self.error(f'"{uuid}" is not a valid UUID')
+
+
+    def check(self, context: Context) -> bool:
+        if self._check_res is True:
```
gkreitz, line 841: Why did you rewrite the memoization of `_check_res` this way? It looks like the memoization is now broken.

problemtools/verifyproblem.py
```diff
@@ -769,73 +773,223 @@ def check(self, context: Context) -> bool:
...
+            return self._check_res
+        elif self._check_res is not False:
+            self._check_res = True
+        to_check = [prop for prop in dir(self) if prop.startswith('_check_') and callable(getattr(self, prop))]
```
gkreitz, line 845: I'm not a fan of trying to find all methods with a special name and running them like this. Why was this approach taken? Do you end up with so many check methods that you can't cleanly list them?

problemtools/verifyproblem.py
```diff
@@ -769,73 +773,223 @@ def check(self, context: Context) -> bool:
...
+        for prop in to_check:
+            self.info(f'Checking "{prop}"')
```
gkreitz, line 847: Internal methods called should not be logged at info level.

problemtools/verifyproblem.py
```diff
@@ -769,73 +773,223 @@ def check(self, context: Context) -> bool:
...
+        best_dist = edit_dist(a, best)
+        for s in b[1:]:
+            dist = edit_dist(a, s)
+            if dist < best_dist:
+                best = s
+                best_dist = dist
+        return best
+
+class ProblemConfigLegacy(ProblemConfigBase):
+    DEFAULT_LIMITS = {
```
gkreitz, line 882: These defaults used to be specified in a `problem.yaml` config file (and user-configurable, as is documented in `README.md`). Does the fact that you list the defaults here imply that you dropped support for having the defaults configurable via config file (and user-configurable via a config file in the home directory)? If so, it feels like a big step backwards to hardcode these values rather than have them in a config file.

problemtools/verifyproblem.py
```diff
@@ -769,73 +773,223 @@ def check(self, context: Context) -> bool:
...
+        "code": 128,
+        "compilation_time": 60,
+        "validation_time": 60,
+        "validation_memory": 1024,
+        "validation_output": 8,
+    }
+
+    def fix_config_structure(self):
+        self._data.setdefault('problem_format_version', formatversion.VERSION_LEGACY)
+        if self._data.get('problem_format_version') != formatversion.VERSION_LEGACY:
```
gkreitz, line 896: These checks (checking that something is one of a few allowed values) get very repetitive. It looks like you ened a helper function, something like
`check_allowed_values(self, key: str, values: list, error_fn = self.error)`

problemtools/verifyproblem.py
```diff
@@ -769,73 +773,223 @@ def check(self, context: Context) -> bool:
...
+            self.error(f'problem_format_version needs to be the string {formatversion.VERSION_LEGACY} for the legacy format (got: "{val}")')
+
+        self._data.setdefault('type', 'pass-fail')
+        if self._data['type'] not in ('pass-fail', 'scoring'):
+            val = self._data['type']
+            self.fatal(f'Type should be one of ["pass-fail", "scoring"] (got: "{val}")')
+
+        if 'name' in self._data:
+            val = self._data['name']
+            if type(val) is not str:
```
gkreitz, line 907: Similarly, here, you should have a `check_type()` function to make the code less repetitive.

Alternatively, both such check functions can probably be implemented using json-schemas.

gkreitz (changes requested):
I started looking at this PR, and I think it needs relatively large amount of work. I don't understand several of the approach changes from #294 at all (not trying to merge the data into a common structure, putting all code in verifyproblem instead of a separate module, not having any test cases).

I ran out of time while reviewing this, but setting this as request changes to not delay giving you the feedback I had for the part I had time to review.

PR #298, opened 2025-04-24
Update problemtools packaging

### LINKED ISSUE #297: Update/rethink packaging of problemtools.
I think we're at a point where we need to re-think packaging for problemtools. The triggering factor is that we'd like to use [nh3](https://github.com/messense/nh3), which is not packaged for the current LTS of Ubuntu, but our current packaging is old and murky in general. It's not quite obvious what to do. I've been trying to read up on things, and my conclusions are:
1. Python packaging of anything but python (and to some extent C extensions) is a pain
2. Debian packaging of python with dependencies is a complete pain

If someone knows more and can point me to good guides for either of the two above, I'd love to hear so. Information I've been able to find is scattered, and typically not up-to-date. Python and deb packaging are not areas I have any expertise in, so there may well be solutions I'm missing.

After some thought, the approach I'm inclined towards is basically:
* Stop trying to coax python builds into running Make and including binaries written in other languages in the python package (side-stepping pain point 1 above). Create two .deb packages - `problemtools` and `problemtools-support`, where the former depends on the latter. `problemtools-support` would contain the binaries from the support directory (possibly minus [checktestdata](https://github.com/DOMjudge/checktestdata), who have their own debian packaging, so we could depend on that too)
* For deb packages, package using `dh-virtualenv` to deploy into a venv, allowing us to have normal pip dependencies and not be locked to OS packages.

This doesn't feel ideal in any way. Bullet point 1 means that just running a `pip install` doesn't give you a working installation (as you really do need the support tools for many use cases). OTOH, to get a working installation from `pip install` today, one must first copy-paste a command line from our README.

Bullet point 2 feels scary as  [dh-virtualenv](https://github.com/spotify/dh-virtualenv) is a pretty dead repo (last commit 3 years ago), so it's unclear for how long the solution will work. I'm not aware of anything better though (see my point 2 above...)

I'll probably experiment a bit and try to do *something* here, but as indicated above, I haven't figured out a solution I feel happy with. So, I'd love to hear if someone else either has opinions/suggestions/pointers to documentation. I'm also slightly curious, we currently list 4 installation approaches (install via `pip`, run via `docker`, run from the repository, or build and install a `deb`) - what are people actually using?

gkreitz:
This PR changes how we package problemtools, to allow us to add pip dependencies without waiting for Ubuntu to package them.

Key changes:

- Build the deb package using `dh_virtualenv` (we now install to `/opt/venvs/kattis-problemtools/` and add symlinks from `/usr/bin` for our tools)
- Clean up python packaging to use the more modern `pyproject.toml`. We still need to keep `setup.py` to fetch and build our support tools.
- Pull version information from git using `setuptools-scm` (this yields a string which is *very* similar to our old versioning)
- Stop exposing `problemtools.__version__`. That PEP failed and was replaced by `importlib.metadata.version`
- Split CI/CD into two separate jobs. Python unit tests now run in parallel with deb building, and deb building only happens once (I don't see a need to test deb building using both pypy and python3)

Fixes #297


README.md
```diff
@@ -31,20 +31,15 @@ A few examples of problem packages can be found in [examples](examples).
 There are four supported ways of installing and running problemtools.
 (For non-Linux users, "Method 2" below, to use Docker, is probably the least painful.)
 
-### Method 1: Install the Python package
+### Method 1: Install the Python package using pipx
 
 Run
 ```
-sudo pip3 install git+https://github.com/kattis/problemtools
+pipx install git+https://github.com/kattis/problemtools
```
pehrsoderman, line 38: Maybe comment where to get pipx?

gkreitz: That's distribution-specific though? And on ubuntu (which feels like the primary distribution to add an instruction for), if you type this without pipx installed, I'm almost certain it'll tell you to `apt install pipx`)

pehrsoderman: I was thinking of dropping a link to https://pipx.pypa.io/stable/installation/ for those not yet familiar with pipx.

gkreitz: Yeah, that's probably a good idea. Added.

PR #311, opened 2025-05-16
Refactor statement handling

### LINKED ISSUE #282: Error if problem name is empty
It should be an error if a problem has an empty (i.e. empty string or only non-printable characters) name.

This is much more likely now due to the changes in the problem format from legacy. In the new format you can use `\problemname` to insert the problem name in the statement, in the old this will set the name to empty string.

### LINKED ISSUE #310: Support for problem_statement latex version "0.1" seems broken (and I will remove it, instead of repairing)
While working on a refactor of our code for problem statements, I tested rendering the ancient format called "0.1" (before `\problemname` was a thing). As far as I can tell, rendering that format is broken and just renders an empty page. As it was replaced around 2012, I don't see much point in debugging further, so I will just remove support for it.

In case anyone actually has a need to render problems that old, I'm adding this issue for some (minor) added visibility. Note that there is an ancient script `problemtools/update_from_old_problemformat.py` which attempts to update problems for you.

gkreitz:
We used to have at least 3 implementations of code that found statements for a problem, all of them inconsistent. This makes use of the new methods in `statement_util` to unify the implementation. That fix blew up a bit, as it touched a lot of old parts (including our latex templates).

I opted not to maintain the "0.1" version of our latex templates any longer. Support for that is now removed. It seemed broken (rendering empty pdf:s) when I tested.

This PR also continues moving from `str` to `pathlib.Path` for paths. A *lot* of our code deals with paths, and I think  using `Path` throughout the code is a significant improvement.

Improves our `mypy` configuration to not allow missing imports. As we are heavy users of a library without type annotations (`plasTeX`), I also made `mypy` follow untyped imports. It seems like `plasTeX` actually has a decent number of type annotations, just that they haven't marked the package as such.

Fixes an untracked bug where `verifyproblem` stopped after the first problem when one listed multiple problems.
Fixes #282
Closes #310


problemtools/tex2html.py
```diff
@@ -2,40 +2,33 @@
...
 
         # Setup parser and renderer etc
-
-        # plasTeX version 3 changed the name of this argument (and guarding against this
-        # by checking plasTeX.__version__ fails on plastex v3.0 which failed to update
-        # __version__)
-        try:
-            tex = plasTeX.TeX.TeX(myfile=texfile)
-        except Exception:
-            tex = plasTeX.TeX.TeX(file=texfile)
```
Matistjati, line 38: Are we sure we want to remove this hack? Requirements.txt requires plasTeX>=3.0, so isn't there a still a risk that you end up with a broken plasTeX version? (I'm not deep in the weeds in Python packing, so I don't know)

problemtools/verifyproblem.py
```diff
@@ -767,54 +760,50 @@ def check(self, context: Context) -> bool:
...
-                options.nopdf = True
-                options.quiet = True
-                if not problem2pdf.convert(options):
-                    langparam = f' --language {lang}' if lang != '' else ''
+            if lang not in self.problem.getMetadata().name:
+                self.error(f'No problem name given in language {lang}')
+            elif not self.problem.getMetadata().name[lang]:
+                self.error(f'Problem name in language {lang} is empty')
+            elif not self.problem.getMetadata().name[lang].strip():
+                self.error(f'Problem name in language {lang} contains only whitespace')
```
Matistjati, line 772: We should probably check that the shortname is valid. That would be implied by the regex "^[a-zA-Z0-9][a-zA-Z0-9_.-]{0,253}[a-zA-Z0-9]$" from https://www.kattis.com/problem-package-format/spec/2023-07-draft.html#general-requirements

Probably better to leave checking that for a follow-up PR, but it really should be done

Matistjati (approved):

problemtools/tex2html.py
gkreitz: As we require plasTeX>=3.0, we're safe and no longer need this hack (it's for supporting <3.0)

problemtools/verifyproblem.py
gkreitz: We already check shortname on https://github.com/Kattis/problemtools/blob/develop/problemtools/verifyproblem.py#L1872 with a stricter regexp: `'^[a-z0-9]+$'`

problemtools/md2html.py
```diff
@@ -154,16 +134,14 @@ def attribute_filter(tag, attribute, value):
...
+        statement_dir: the directory with problem statement files
         img_src: the image source as in the Markdown statement
     """
 
-    source_name = os.path.join(problem_root, 'statement', img_src)
-
     if os.path.isfile(img_src):  # already copied
         return
-    shutil.copyfile(source_name, img_src)
+    shutil.copyfile(statement_dir / img_src, img_src)
```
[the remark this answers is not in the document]
gkreitz: Good point! I'll leave that for a follow-up

problemtools/md2html.py
```diff
@@ -14,28 +14,19 @@
 from . import statement_util
 
 
-def convert(problem: str, options: argparse.Namespace) -> bool:
+def convert(problem_root: Path, options: argparse.Namespace, statement_file: Path) -> bool:
     """Convert a Markdown statement to HTML
```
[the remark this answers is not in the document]
gkreitz: Will it? I'm pretty sure it'll be written to `destfile`.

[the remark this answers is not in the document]
gkreitz: Good point. Feels like that also goes into the follow-up that deals with the other issue you raised w.r.t. copy_image.

PR #315, opened 2025-05-16
Replace formatversion.FormatData with a StrEnum

gkreitz:
The earlier code was a bit annoying to use. This PR replaces `FormatData` with an enum called `FormatVersion` with the same properties as `FormatData` had.

### DIFF
--- a/problemtools/formatversion.py
+++ b/problemtools/formatversion.py
@@ -1,92 +1,47 @@
-import os
 import yaml
-from dataclasses import dataclass
+from enum import StrEnum
+from pathlib import Path
 
 
-VERSION_LEGACY = 'legacy'
-VERSION_2023_07 = '2023-07-draft'
+class FormatVersion(StrEnum):
+    LEGACY = 'legacy'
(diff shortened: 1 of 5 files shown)


problemtools/formatversion.py
```diff
@@ -1,92 +1,47 @@
...
+class FormatVersion(StrEnum):
+    LEGACY = 'legacy'
+    V_2023_07 = '2023-07-draft'  # When 2023-07 is finalized, replace this and update _missing_
+
+    @property
+    def statement_directory(self) -> str:
+        match self:
+            case FormatVersion.LEGACY:
+                return 'problem_statement'
+            case FormatVersion.V_2023_07:
```
pehrsoderman, line 15: I am not 100% on the syntax here, but I would expect a default to throw an exception?

pehrsoderman (approved):

problemtools/formatversion.py
gkreitz: `mypy` catches if we miss a case and complains about a missing return statement, so I prefer not to clutter the code with extra lines as we already have a tool saving us.

PR #318, opened 2025-05-19
Improve image handling in markdown statements

gkreitz:
* Support images in subdirectories (`statement/img/foo.png` linked as `img/foo.png`)
 * Add more security-focused checks on what `img` `src` tags we allow

Thanks to @Matistjati for finding these errors

### DIFF
--- a/tests/problems/imgrequest2/statement/problem.en.md
+++ b/tests/problems/imgrequest2/statement/problem.en.md
@@ -1,3 +1,3 @@
 Make web request via image
 
-<img src="http:picsum.photos/400">
+<img src="https:open.kattis.com/images/site/header/logo-empty.png">
--- a/problemtools/md2html.py
+++ b/problemtools/md2html.py
@@ -1,6 +1,7 @@
 #! /usr/bin/env python3
 # -*- coding: utf-8 -*-
 import argparse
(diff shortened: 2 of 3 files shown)


pehrsoderman (approved):

tests/problems/imgrequest2/statement/problem.en.md
```diff
@@ -1,3 +1,3 @@
 Make web request via image
 
-<img src="http:picsum.photos/400">
+<img src="https:open.kattis.com/images/site/header/logo-empty.png">
```
simonlindholm, line 3: What's up with the lack of `//` here?

gkreitz: This tests for a (mild) vulnerability that existed previously where a markdown statement could link to an external image (per the text two lines above). The link `https:open.kattis.com/images/site/header/logo-empty.png` will be interpreted by a browser as a normal https link, as if the `//` was there. However, it's also a valid relative path on a unix system, and that files exists in a subdirectory to this file.

We want to make sure images are all contained in problem packages and that the rendered markdown does not allow links to external sites (even if the problem packager is malicious and drops a suitably named file).

PR #320, opened 2025-05-22
Restore most of the old API offered by `problemtools.verifyproblem.Problem`

gkreitz:
We don't really offer a stable API, but this restores most of the old API used to access aspects of a problem (config remaining a large, breaking change).
    
The rework and API changes that had been done early in the process of  supporting 2023-07-draft do not seem to be needed, so this PR reverts much of it to simplify `verifyproblem` and ease porting API users (mostly Kattis, I assume) to the new version by keeping the number of breaking changes down.

Adds two minor changes to the API (and keeps the large change that metadata is now accessed via `problem.metadata` instead of `problem.config`):

- The type of `problems.attachments.attachments` changed from `list[str]` to `list[Path]`
- To access the computed time limit, use `problem.timelim` (previously, it was put into `config`)

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -1663,7 +1663,7 @@ class Submissions(ProblemPart):
                     max_runtime = max(runtimes)
                     exact_timelim = max_runtime * time_multiplier
                     max_runtime_str = f'{max_runtime:.3f}'
-                    timelim = max(1, int(0.5 + exact_timelim))
+                    timelim = max(1, int(0.5 + exact_timelim))  # TODO: properly support 2023-07 time limit computation
                     timelim_margin_lo = max(1, min(int(0.5 + exact_timelim / safety_margin), timelim - 1))
                     timelim_margin = max(timelim + 1, int(0.5 + exact_timelim * safety_margin))
                 else:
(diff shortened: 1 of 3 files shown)


problemtools/verifyproblem.py
```diff
@@ -232,16 +215,16 @@ def check(self, context: Context) -> bool:
 class TestCase(ProblemAspect):
     Result = tuple[SubmissionResult, SubmissionResult, SubmissionResult]
 
-    def __init__(self, problem: Problem, aspect_name: str, base: str, testcasegroup: TestCaseGroup) -> None:
-        super().__init__(f'{problem.shortname}.{aspect_name}.{testcasegroup.name}.{os.path.basename(base)}', problem)
+    def __init__(self, problem: Problem, base: str, testcasegroup: TestCaseGroup) -> None:
+        super().__init__(f'{problem.shortname}.test.{testcasegroup.name}.{os.path.basename(base)}', problem)
```
pehrsoderman, line 219: Why is `test` needed here?

gkreitz: It's not *needed*, it's just the old string that was there earlier. These strings are passed to `logging` to fetch the logger, so we want a hierarchy, so we want * something* in between shortname and testcasegroup.name to denote the test cases. There are tools which apparently parse verifyproblem output https://github.com/Kodsport/testdata_tools/blob/master/analyzetestgroups.py (yikes!), so I figured just sticking the string as it was in the earlier release made the most sense.

PR #326, opened 2025-05-27
Run verifyproblem on all examples in CI

### LINKED ISSUE #325: Run `verifyproblem` on all examples in CI
Our CI should run `verifyproblem` on all example problems to help catch errors. While we *could* technically just add it to our workflow scripts right now, installing all required languages to successfully verify examples takes quite a while (5-10 minutes, installing just shy of 1000 deb packages...)

Thus, my plan is to build a CI docker image (#321) which is just an ubuntu with all the requisite `.deb` packages to build and run problemtools (with all supported languages) and use that in our workflows. The image ends up quite large (6GB), so we'll see what performance is like.

gkreitz:
Looks like unfortunately this will need to stop testing with pypy in CI, as pip install of nh3 fails with:
```
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [4 lines of output]
      Checking for Rust toolchain....
      Rust not found, installing into a temporary directory
      Python reports SOABI: pypy39-pp73
      Unsupported platform: pp73
```
`nh3` does not seem to provide binary wheels for pypy, and I gave up on trying to debug that build issue.  Instead of pypy, we now test with CPython 3.11, 3.12, and 3.13 to get better coverage there. I'm guessing the vast majority of users will use CPython to run problemtools anyway.

Fixes #325

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -17,7 +17,9 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["pypy3.11", "3.11"] # 3.11 is the lowest we support, since we want StrEnum
+        python-version: ["3.11", "3.12", "3.13"] # 3.11 is the lowest we support, since we want StrEnum
+    container:
+      image: problemtools/githubci:latest
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/python-app.yml
```diff
@@ -26,29 +28,35 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip
-        pip install mypy ruff pytest
-        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
-        sudo apt-get update
-        sudo apt-get install pandoc tidy ghostscript python3 texlive-fonts-recommended texlive-lang-cyrillic texlive-latex-extra texlive-plain-generic
+        python -m venv venv
+        venv/bin/python --version
```
pehrsoderman, line 32: Leftover debugging?

pehrsoderman:
Do you want me to remove the requirement to build on pypy?

.github/workflows/python-app.yml
gkreitz: Debugging yes. Not quite leftover though, I felt it was sort of reasonable to leave behind (costs nothing, and may come in handy again in the future).

PR #347, opened 2025-08-16
Fuzz input validators for trailing whitespace

Matistjati:
At the time of writing, 16 Aug 2025, both https://open.kattis.com/problems/procrastination and https://open.kattis.com/problems/ekkiminnforseti have trailing whitespaces in their samples. This is very minor, but should IMO not happen.

This PR adds tests that check the input validator will reject this.

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -975,6 +975,7 @@ _JUNK_MODIFICATIONS = [
     _build_junk_modifier(
         'spaces added where there already is whitespace', r'\s', lambda m: m.group(0) + ' ' * random.randint(1, 5)
     ),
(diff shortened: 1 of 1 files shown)


Matistjati:
On another note, as far as I can tell, we use `random` without seeding. This seems like it could be extremely annoying in some edge-case. Do you see any problems in naively adding `random.seed(42)` to the top of `verifyproblem.py` @gkreitz ?

gkreitz (approved):

gkreitz:
While I added a comment about how we do random, I think this is a very good addition by itself, so I'll merge it now, and then we can do a separate PR if we want to adjust randomness per my comment above (explicit seeding in a few places, and just adding a single space/newline in a couple of cases).

PR #348, opened 2025-08-19
Fuzz output validators

Matistjati:
I suspect that many output validators out there are not crash-proof against all contestant output. I think that adding a couple of evil testcases might catch some poorly written ones.

I decided to only run it on samples, since performance actually becomes a concern if you have lots of testcases (especially if they don't deduplicate symlinks. Not sure if we do this, will have to check in the future).

For the contents of the cases, I just tried to think of ways I to incorrectly write some existing validators. Perhaps @niemela or @gkreitz could look into judge errors on Kattis and add a minimal version of what killed the output validator?

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -1318,7 +1318,11 @@ class OutputValidators(ProblemPart):
 
             # For performance reasons, strongly limit the amount of testcases we run on
             fast_languages = {'c', 'cpp'}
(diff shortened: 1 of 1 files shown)


Matistjati:
> I think a hard coded list of test cases like this is likely a bit too simplistic to yield good results though, but it's a simple enough addition to add, and if it helps catch one or a few bugs, it's worth it.
You'd be surprised to the amount of new authors who write output validators that will blow up if you sneeze on them. I've seen this at least 5 times.

PR #349, opened 2025-08-19
Deterministic input fuzzing

Matistjati:
As a follow-up to https://github.com/Kattis/problemtools/pull/347, make input fuzzing deterministic.

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -960,7 +960,7 @@ _JUNK_CASES = [
     ('a text file with the ASCII characters 32 up to 127', bytearray(x for x in range(32, 127))),
     (
         'a random text file with printable ASCII characters',
-        (lambda rng=random.Random(42): bytearray(rng.choice(string.printable.encode('utf8')) for _ in range(200)))(),
+        (lambda rng: bytearray(rng.choice(string.printable.encode('utf8')) for _ in range(200)))(random.Random(42)),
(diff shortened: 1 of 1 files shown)


Matistjati:
We should probably merge https://github.com/Kattis/problemtools/pull/348 first. I based this PR off that one. I assumed Github would do something smart and only show the new changes w.r.t. the last one, but upon further thoughts, I don't know what I expected, really.

problemtools/verifyproblem.py
```diff
@@ -971,18 +993,17 @@ def _build_junk_modifier(
     return (desc, p.search, lambda text: p.sub(repl, text))
 
 
+rng = random.Random(42)
```
gkreitz, line 996: I would prefer not creating a variable `rng` for this, but instead do `random.Random(42).choice(...)` down where you use this variable.

If we create `rng` and multiple modifiers start using it, we lose determinism if they are reordered or removed (probably no big deal, but if the point of this PR is determinism... :)

gkreitz: Oh, good catch. I didn't think about the context with how it's called.

Matistjati:
Uhh... accidentally deleted this branch when cleaning up old branches because I thought it was merged. Whoops.

Matistjati:
Closed; see https://github.com/Kattis/problemtools/pull/351 instead

PR #350, opened 2025-08-20
Multipass support

Matistjati:
I dislike the inline functions on line 415 and 421, but ruff demands it...

I do not try to prevent the submission from passing information out of bound. In my opinion, doing so would just be cosmetic, as it's impossible to do it fully without building a sandbox.

In a follow-up PR, I will fix multipass sample rendering when the source is tex. I have fixed it for md -> html in this PR.

### DIFF
--- a/problemtools/statement_util.py
+++ b/problemtools/statement_util.py
@@ -5,12 +5,15 @@ import os
 import re
 import subprocess
 import tempfile
+import logging
 from pathlib import Path
 from typing import Optional, List, Tuple
 from urllib.parse import urlparse
 
 from . import metadata
 from .formatversion import FormatVersion, get_format_version
+
+log = logging.getLogger(__name__)
 
 ALLOWED_IMAGE_EXTENSIONS = ('.png', '.jpg', '.jpeg')  # ".svg"
 FOOTNOTES_STRINGS = ['<section class="footnotes">', '<aside class="footnotes">']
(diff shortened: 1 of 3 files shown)


Matistjati:
I will split up the tex -> html/pdf into another PR. This PR is ready for review now

problemtools/statement_util.py
```diff
@@ -235,39 +246,87 @@ def format_normal_sample(sample_root: str, sample: str, casenum: int) -> str:
...
-            line_type = 'sampleinteractionwrite'
-        elif interaction[0] == '<':
-            line_type = 'sampleinteractionread'
-        else:
-            print(f'Warning: Interaction had unknown prefix {interaction[0]}')
-        lines.append(f"""<div class="{line_type}"><pre>{data}</pre></div>""")
+        if interaction.startswith('---'):
+            passes.append(curr_pass)
+            curr_pass = []
+            continue
```
gkreitz, line 318: This `continue` could probably be an `else` (and I think that'd be preferable)

problemtools/verifyproblem.py
```diff
@@ -1376,15 +1436,21 @@ def _parse_validator_results(self, val, status: int, feedbackdir, testcase: Test
...
+        if self.problem.metadata.is_custom_score_mandatory():
             if os.path.isfile(score_file):
                 try:
                     score_str = open(score_file).read()
                     score = float(score_str)
                 except Exception as e:
                     return SubmissionResult('JE', reason=f'failed to parse validator score: {e}')
             else:
-                return SubmissionResult('JE', reason='problem has custom scoring but validator did not produce "score.txt"')
+                # If we're running multipass, we do not need to output a score after every pass
```
gkreitz, line 1447: Do you actually need this check? Multi-pass can only happen in the new format, where custom score is never mandatory. So I think you could have left this else-clause as-is?

(If we do in fact need it, can't we just check both that the problem is multipass *and* that nextpass.in exists, making the check correct and avoiding a scary comment about accepting a small risk? :)

gkreitz (approved):
Looks good. Added two nit-picky comments, but they are so small I'll just merge this. We can fix those in a future PR if we want/care.

problemtools/verifyproblem.py
Matistjati: Yes, that's true. I somehow missed that we actually have a reference to the metadata via self.problem in this class. Will add both of these in the multipass sample PR.

PR #370, opened 2025-11-27
Allow underscore as start of name of files and directories.

pehrsoderman:


gkreitz:
We should probably just take the regex from https://www.kattis.com/problem-package-format/spec/2025-09.html#general-requirements for file and directory names instead of the change here.

Nice find with _SUB_REGEXP. I think that should most likely be removed, or perhaps be changed to something like `^[^.].*` if we want to filter out things with leading periods (but we probably should deal with that in a more general way than via this regexp). I don't think we want to maintain a separate (basically identical, plus allowing the archaic `c++` suffix :)) regex there.

pehrsoderman:
Replaced by #371

PR #376, opened 2025-12-06
Better grader error display

Matistjati:
This PR does 3 things:
- Rewrites some awkward `open()` code to be crash-safe
- Prints stderr in case the grader crashes
- Removes support for multiple graders

Sample outputs:

Differing grader outputs:
```
ERROR Different graders gave different results:
ERROR grader.py: AC 0.0
ERROR grader2.py: AC 1.0
```

Grader stderr (bizarre error, I know):
```
Checking submissions
ERROR Judge error: grader.py (Python 3 (w/PyPy3)) crashed
ERROR Grader stderr:
==134096==ERROR: AddressSanitizer failed to allocate 0xdfff0001000 (15392894357504) bytes at address 2008fff7000 (errno: 12)
==134096==ReserveShadowMemoryRange failed while trying to map 0xdfff0001000 bytes. Perhaps you're using ulimit -v


ERROR Judge error: grader.py (Python 3 (w/PyPy3)) crashed
ERROR Grader stderr:
==134097==ERROR: AddressSanitizer failed to allocate 0xdfff0001000 (15392894357504) bytes at address 2008fff7000 (errno: 12)
==134097==ReserveShadowMemoryRange failed while trying to map 0xdfff0001000 bytes. Perhaps you're using ulimit -v
...
```

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -1274,7 +1274,7 @@ class Graders(ProblemPart):
         score: float = 0
 
         if not sub_results:
-            self.info('No results on %s, so no grader ran' % (testcasegroup,))
+            self.info(f'No results on {testcasegroup}, so no grader ran')
             return (verdict, score)
 
         grader_flags = testcasegroup.config['grader_flags'].split()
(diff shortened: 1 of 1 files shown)


gkreitz (approved):

Matistjati (commented):
Me and Fredrik agree that there should be at most one grader, and that we should add it to the spec. We decided not to rename the folder to `grader` for backwards compatibility, and thus I didn't rename the class name.

PR #385, opened 2026-01-05
Fix missing header in default validator

evouga:
I needed \<tuple\> for std::tie to compile (on g++ 9.4.0).


gkreitz:
Thank you!

PR #405, opened 2026-04-05
Set cwd when running input validators

### LINKED ISSUE #336: Verifying `examples/oddecho` drops a core in the working directory
I haven't debugged this fully, but we do run the input validator on something where it does an `assert` while the working directory is the current directory, causing a core to be dumped. Commenting out the `assert` from the input validator stops the core from being dumped.

I'm guessing the problem is when we run junk cases, where we do not seem to set a working directory.

Steps to reproduce:
```sh
ulimit -c unlimited
bin/verifyproblem.sh -p validators -- examples/oddecho
# Also reproducible with bin/verifyproblem.sh -p data -- examples/oddecho
ls core
```

Matistjati:
Closes #336. I think that this is the right solution- changing the limit ourselves adds more noise to the codebase as opposed to just masking it by dumping the core in the temporary working directory.

Doesn't result in a statistically significant slowdown.
```python
matistjati@DESKTOP-BJV27M6:~/po/swedish-olympiad-2026/lager2$ ulimit -c unlimited
matistjati@DESKTOP-BJV27M6:~/po/swedish-olympiad-2026/lager2$ time ~/kattis/problemtools/bin/verifyproblem.sh fredrikspizzeria -p data
Loading problem fredrikspizzeria
Checking data
fredrikspizzeria tested: 0 errors, 0 warnings

real    0m25.040s
user    0m20.809s
sys     0m2.376s
matistjati@DESKTOP-BJV27M6:~/po/swedish-olympiad-2026/lager2$ ulimit -c 0
matistjati@DESKTOP-BJV27M6:~/po/swedish-olympiad-2026/lager2$ time ~/kattis/problemtools/bin/verifyproblem.sh fredrikspizzeria -p data
Loading problem fredrikspizzeria
Checking data
fredrikspizzeria tested: 0 errors, 0 warnings

real    0m24.472s
user    0m20.102s
sys     0m2.428s
```

### DIFF
--- a/problemtools/verifyproblem.py
+++ b/problemtools/verifyproblem.py
@@ -1158,7 +1158,7 @@ class InputValidators(ProblemPart):
                 for flags_str in all_flags:
                     flags = flags_str.split()
                     for val in self._validators:
-                        status, _ = val.run(file_name, args=flags)
+                        status, _ = val.run(file_name, args=flags, work_dir=self.problem.tmpdir)
                         if os.WEXITSTATUS(status) != 42:
                             break
                     else:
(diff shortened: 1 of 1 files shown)


gkreitz (commented):
Ideally, we should probably try to follow https://www.kattis.com/problem-package-format/spec/2025-09.html#working-directory here (even if we need to fudge things a bit - as there isn't really a submission here...)

But it's clear progress not to litter the package with core files, thanks! :)

gkreitz (approved):

Issue #398, opened 2026-03-19
Refactor `verifyproblem.py`


gkreitz:
A *lot* of the functionality of this tool sits in a single python file, `verifyproblem.py` which has grown organically over the years. It's now 2.2k+ lines, and I feel like it's no longer very readable or understandable, and that's hindering progress.

I'm adding this issue mostly for visibility, to acknowledge the problem, and to indicate I'm working on it (albeit not super focused). If you have opinions on the architecture, feel free to chime in.

My current inclination (subject to change) is to head in a direction where we roughly have:

- `problemtools.model` - a module with dataclasses (or similar) of a parsed problem. These would liberally attempt to load a problem and its constituent parts, and have as little check logic as possible
- `problemtools.checks` - a module with the checks from verifyproblem (not all in one file :))
- `verifyproblem.py` - a smallish program loading the model, running the checks, and reporting the result

Sketching `problemtools.model`, I'm thinking something *very* roughly like this:
```
# Dataclasses for the individual parts
@dataclass
class Statement:
    format: Literal['latex', 'markdown]
    language: str
    file: Path
...

@dataclass
class Problem:
     statements: list[Statement]
     config: ProblemConfig
     ...

def _load_statements(...) -> list[Statement]
def _load_config(...) -> ProblemConfig
...

def load_problem(path, diagnostics) -> Problem, Diagnostics, list[Path]  # unknown files in the package
```

Diagnostics is a new class to which warnings, notices, and errors are logged. I envision that one may either want an implementation which just captures everything in-memory for API usage, or something that also logs messages with color, as they happen.

Matistjati:
I might be overengineering, but it would be nice to have submissions result be dataclasses, and have callbacks with the partially-filled-in results before and after every testcase run. This way, creating different views of the output would be easy. In principle, you could even build an online judge using it (although that would be kind of risky due to a lack of sandboxing).

gkreitz:
Brief update, as I've had some time to put towards this. I've added the Diagnostics part, as that was a step I felt reasonably confident in, and wasn't a huge PR. I've been prototyping a bit, and I think the direction I envisioned is feasible, and a clear improvement on the current state. I'm struggling a bit to see how to do this in reasonably-sized PR:s, but hopefully I'll come up with something.

Basic structure I'm looking at now is still very close to the above. Field types are very much prototype, I've been playing around a bit with different ideas.
```python
# models/problem.py
@dataclass
class Problem:
    probdir: Path
[...]

Issue #382, opened 2025-12-22
Better output for scoring problems


Matistjati:
@thorehusfeldt has created a script `analyzetestgroups.py`, which creates beautiful output for scoring problems. (Available here https://github.com/Kodsport/testdata_tools/blob/master/analyzetestgroups.py)

<img width="985" height="247" alt="Image" src="https://github.com/user-attachments/assets/01e40e84-14f2-4d53-9f39-8f1aeedaaaa7" />

We should consider integrating this into verifyproblem for scoring problems.

Matistjati:
I got an LLM to implement this in my fork. Use at your own risk https://github.com/Matistjati/problemtools/tree/llm-additions. While the code change itself is not super large (about 200 LOC), this is probably blocked by #398 . Additionally, it's not entirely clear where this fits in. I think it would be best to let verifyproblem have two modes: visualize and debug, where debug gives a maximal amount of information (slowest case, first failing case, etc.), and visualize would give this table for scoring problems.

Matistjati:
Some inspiration on how it can look like

<img width="1842" height="399" alt="Image" src="https://github.com/user-attachments/assets/dfbb28ac-aef7-4f1a-b608-7f189a507f41" />

It's implemented using the python package rich. I'm not completely sure how we want to display the "sensitive to time limit" texts. It also doesn't display slowest test case per submission.

Issue #379, opened 2025-12-22
Add support for caching verdicts


Matistjati:
Basically, add a command-line option that when activated, on every testcase run, hashes the testcase, grader_flags, grader, output validator and output validator flags, and submission. 

This is useful in problems with slow solutions, lots of testdata, where you want to add testcases and see if the verdict of some given submissions change.

gkreitz:
I'd forgotten about this issue. By happy accident, in #398, we're going to get much closer to being able to support this. Internally while running, we'll actually have an in-RAM cache almost exactly like this (with with a simpler representation of the sub and validator, as those don't change during a run).

There is one annoying quirk with a cache like this - time limits. If you know a submission times out with some config in X seconds and the test data has now changed so that the time limit is > X, that cache entry is suddenly worthless (because there's always the chance that the submission would now do something different).

gkreitz:
It would also be nice to cache the success of statement rendering, as that takes a while.

PR #411, opened 2026-04-27
Extract all logic for running and judging a submission from verifyproblem to judge module #398

### LINKED ISSUE #244: Symlinks for reuse of input files with different output validator flags

### LINKED ISSUE #383: We leak non-integer time limits
Output from running on latest version
```
   Slowest AC runtime: 5.115, setting timelim to 7 secs, safety margin to 14 secs
   PAC submission decreasing_priorities_UFQ.cpp (C++) OK: AC (26) [testcase: testcase secret/g7full/003-g1-3, CPU: 2.46s @ testcase secret/g6decreasing/076-g6-3]
   PAC submission full_solution.py (Python 3 w/Pypy) OK: AC (19) [testcase: testcase secret/g7full/037-g3-1, CPU: 21.15s @ testcase secret/g4line/055-g4-3]
   PAC submission increasing_priorities_UF_undo.cpp (C++) OK: AC (14) [testcase: testcase secret/g7full/002-g1-2, CPU: 2.54s @ testcase secret/g5increasing/066-g5-2]
   PAC submission line_st.cpp (C++) OK: AC (17) [testcase: testcase secret/g7full/001-g1-1, CPU: 2.54s @ testcase secret/g4line/055-g4-3]
WARNING PAC submission line_st.py (Python 3 w/Pypy) sensitive to time limit: limit of 5.833333333333334 secs -> WA, limit of 14.0 secs -> AC
```

Something seems a little strange to me: a fractional time limit of `5.833333333333334` and computed time limit of `7 secs` don't seem consistent. So there is probably also a bug hiding somewhere.

### LINKED ISSUE #397: Behavior on empty testgroups
A common scenario is that I want to run a particular solution on only one testgroup to find the first failing testcase in said testgroup. I would then run something like
`verifyproblem <problem> -s <submission> -t 1 -d secret/group1`
If the submission then gets wrong answer on group1, I would expect the overall verdict to be WA. However, because empty groups give AC (0), the overall verdict is 0.
Personally, I would prefer for the verdict for empty groups to the identity element (at least conceptually). Should not be hard: the code already detects that the groups are empty (ran with `-l debug`):
```python
INFO Running on testcase group data.secret.group2
INFO No results on testcase group data.secret.group2, so no grader ran
INFO No results on testcase group data.secret.group2, so no grader ran
INFO No results on testcase group data.secret.group2, so no grader ran
INFO Running on testcase group data.secret.group3
INFO No results on testcase group data.secret.group3, so no grader ran
INFO No results on testcase group data.secret.group3, so no grader ran
INFO No results on testcase group data.secret.group3, so no grader ran
INFO Running on testcase group data.secret.group4
INFO No results on testcase group data.secret.group4, so no grader ran
INFO No results on testcase group data.secret.group4, so no grader ran
INFO No results on testcase group data.secret.group4, so no grader ran
INFO Running on testcase group data.secret.group5
INFO No results on testcase group data.secret.group5, so no grader ran
INFO No results on testcase group data.secret.group5, so no grader ran
INFO No results on testcase group data.secret.group5, so no grader ran
INFO Running on testcase group data.secret.group6
INFO No results on testcase group data.secret.group6, so no grader ran
INFO No results on testcase group data.secret.group6, so no grader ran
INFO No results on testcase group data.secret.group6, so no grader ran
DEBUG Grading 6 results:
WA 0.0
AC 0
AC 0
AC 0
AC 0
AC 0
```

gkreitz:
This PR ended up huge, but I don't think there was any way to avoid that.

This is a reimplementation of the logic for running a submission, moving everything from `verifyproblem` to the module `problemtools.judge`. Earlier code was quite confusing (IMHO), where it kept passing round 3 time limits, and returned the results for those 3 different time limits. The basic idea was good though - if we know what happened at some time limit t, we can cheaply compute what happens for t* <= t by toggling test cases to TLE and running the grader.

In the new implementation, the core piece is the `ResultStore` class, which achieves 3 things:

- Reuse results for identical test cases - the `ResultStore` is a cache, and the cache key is what determines if we can reuse a result (so, sha256 of input, output, validator flags, ...).
- Deal with multithreading - workers claim entries into the cache, and readers can get a future back when looking up items
- Implements simulating what would happen at lower time limits.

The `ResultStore` is currently instantiated per submission, but we'll probably want to extend this later and add some persistence, cf. #379.

Test case reuse logic is now much improved, and symlinks between test cases are no longer a magical way to enable result reuse. If two test cases are identical, we will reuse results (with or without symlinks). It is also perfectly legal to add a symlink to save space when files are identical, even if the entire test case isn't.

The main interface to the `judge` module is the class `SubmissionJudge` which knows how to judge a submission. Instead of just returning an aggregated result, it returns all intermediary results for the entire test case tree. This allows `SubmissionResult` to be a much simpler class, as the caller can easily compute various warnings (e.g., failing on sample cases) based on the tree instead of needing that information to be aggregated.

Fixes #397 
Fixes #383 
Progress on #382 
Progress on #379 
Fixes #244

### DIFF
--- a/problemtools/judge/__init__.py
+++ b/problemtools/judge/__init__.py
@@ -1,17 +1,19 @@
+from .cache import CacheKey
 from .execute import execute_testcase
 from .result import (
     SubmissionResult,
     TimeLimits,
     Verdict,
-    classify_result,
 )
+from .submission_judge import SubmissionJudge
 from .validate import validate_output
 
 __all__ = [
+    'CacheKey',
+    'SubmissionJudge',
     'SubmissionResult',
     'TimeLimits',
     'Verdict',
-    'classify_result',
     'execute_testcase',
     'validate_output',
 ]
--- a/problemtools/judge/__init__.py
+++ b/problemtools/judge/__init__.py
@@ -2,7 +2,6 @@ from .cache import CacheKey
 from .execute import execute_testcase
 from .result import (
     SubmissionResult,
-    TimeLimits,
     Verdict,
 )
 from .submission_judge import SubmissionJudge
@@ -12,7 +11,6 @@ __all__ = [
     'CacheKey',
     'SubmissionJudge',
     'SubmissionResult',
-    'TimeLimits',
     'Verdict',
     'execute_testcase',
     'validate_output',
--- a/problemtools/judge/cache.py
+++ b/problemtools/judge/cache.py
@@ -41,11 +41,12 @@ def _reclassify(result: SubmissionResult, timelim: float) -> SubmissionResult:
 
 
 def _with_test_node(result: SubmissionResult, testcase: TestCase) -> SubmissionResult:
-    """Return result with test_node set to testcase, copying only if needed."""
-    if result.test_node is testcase:
+    """Return result with test_node and runtime_testcase set to testcase, copying only if needed."""
+    if result.test_node is testcase and result.runtime_testcase is testcase:
         return result
     result = copy.copy(result)
     result.test_node = testcase
+    result.runtime_testcase = testcase
     return result
 
 
--- a/problemtools/judge/cache.py
+++ b/problemtools/judge/cache.py
@@ -40,12 +40,12 @@ def _reclassify(result: SubmissionResult, timelim: float) -> SubmissionResult:
     return result
 
 
-def _with_test_item(result: SubmissionResult, testcase: TestCase) -> SubmissionResult:
-    """Return result with test_item set to testcase, copying only if needed."""
-    if result.test_item is testcase:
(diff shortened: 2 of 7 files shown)


pehrsoderman (approved):
Very hard to review, but I can't find any obvious problems.

Issue #103, opened 2018-11-19
languages.yaml should accept compilers in path and not hardwire absolute paths


godmar:
Is there a reason for why languages.yaml hardwires the absolute path of all compilers?

It causes issues when I need to use a particular version of gcc/g++ on systems where /usr/bin/g++ is not the version problemtools wants, such as CentOS.  On those systems, you start a shell such that `/opt/rh/devtoolset-7/root/usr/bin/g++` is in the path before `/usr/bin/g++` using [RedHat's scl-utils](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Developer_Guide/scl-utils.html).

problemtools completely breaks that with its assumption that the [absolute path to an executable be provided in languages.yaml](https://github.com/Kattis/problemtools/blob/develop/problemtools/run/source.py#L100-L104)

Do you want it this way, or would you entertain a PR to accept commands from the path in addition?

austrin:
> Is there a reason for why languages.yaml hardwires the absolute path of all compilers?

Yes, but it's not very good (or it used to be good but has played out, or I forgot what it was, doesn't really matter).

A PR fixing the behaviour you mention that verifyproblem breaks if absolute paths are not provided would definitely be welcome.

Whether to also change the defaults in problem.yaml or not, I'd like to mull over this a little while first, but given the recently added https://github.com/Kattis/problemtools#configuration this should also be relatively easy for you to work around once the first issue has been fixed.

PR #436, opened 2026-06-30
Replace the checktestdata binary with the python reimplementation

mzuenni:


problemtools/run/checktestdata.py
```diff
@@ -3,25 +3,24 @@
...
-    _CTD_PATH = get_tool_path('checktestdata')
-
     def __init__(self, path):
         """Create a Checktestdata wrapper.
 
         Args:
             path (str): path to .ctd source file
         """
-        if Checktestdata._CTD_PATH is None:
+        pyctd = shutil.which('pyctd')
```
gkreitz: I don't think we want to search `$PATH` for tools we ship with, that feels very weird.

As this is a python library, why do we even need to spawn a new process and python interpreter for this? Do you have any plans for a stable API to call? I'd strongly prefer a library integration over running a python tool this way.

gkreitz (changes requested):
I'm not sure we want to integrate `pyctd` as a command-line tool. A large part of the upside of a python tool here would be that we can integrate it in a cleaner way (the way we ship binary tools with problemtools is problematic, and I hope to some day get rid of it).

Do you have any plans for providing a stable API + type annotations (and a `py.typed` marker)?

problemtools/run/checktestdata.py
mzuenni: My main idea was to have `pyctd` as a drop-in replacement for the checktestdata binary (mainly because it should make switching trivial). If you want you could call an api (and I would add one for that) but I am not sure if that should be preferred.
That would need larger changes on your codebase... but the main issue would be that this opens questions regarding sandboxing, timing, GIL and many more (For example, the pyctd application sets `sys.set_int_max_str_digits(0)` which would then leak to problemtools?)

> I don't think we want to search $PATH for tools we ship with, that feels very weird.

As far as I know that is not too uncommon.

gkreitz: > (For example, the pyctd application sets sys.set_int_max_str_digits(0) which would then leak to problemtools?)

Yikes, yes, thanks for the heads-up. If you need to modify global state, then using this as a library is a non-starter.

Perhaps the upside of moving to a python implementation isn't as large as I first anticipated then. I sort of assumed this would become a library, not something written to only be usable as a command-line tool.

I also see now that your project is licensed under GPL, which is a bit of a blocker. I'm honestly unsure how GPL treats this type of usage. The license choice makes library usage problematic at least.

> sandboxing

Currently, problemtools does no sandboxing by itself, so that one is easy. ;) Is there any way to write malicious ctd?

> timing

I must admit, I'm not really up-to-date with the ctd language. Can the user write a program that's slow/hangs? I was under the assumption that it was basically just describing the shape of the data, so performance is what it is.

> GIL

I would have assumed the programs you generate are pretty I/O-bound, so that the GIL isn't really an issue? But relatedly, as a library, thread safety guarantees would be important.

> As far as I know that is not too uncommon.

Can you give some examples of other projects doing this? I can't recall seeing people doing this. I am quite strongly opposed to digging through `$PATH` to find a tool we ourselves ship. I'm guessing the assumption that `pyctd` would be installed in `$PATH` may be what's making this PR fail CI.

I'm honestly unsure what the best practice is for running a command line tool installed as a python dependency. I hope it's not some ugly hack like our `get_tool_path` (that's one of the drawbacks of shipping binaries with a python project, there's just no clean way to do so that I've found - `get_tool_path` is code I would love to evict).

mzuenni: >> (For example, the pyctd application sets sys.set_int_max_str_digits(0) which would then leak to problemtools?)
>
> Yikes, yes, thanks for the heads-up. If you need to modify global state, then using this as a library is a non-starter.

That is the most obvious state (and one that is unavoidable) but also one that wouldn't harm problemtools. I am not sure what other global state leaks in either direction... the subinterpreter is just safer in that regard

> Perhaps the upside of moving to a python implementation isn't as large as I first anticipated then. I sort of assumed this would become a library, not something written to only be usable as a command-line tool.

The upside is still that you do not need to compile anything and get cleaer dependecies. I mostly removed stuff in this pr :)

>> sandboxing
>
> Currently, problemtools does no sandboxing by itself, so that one is easy. ;) Is there any way to write malicious ctd?

no. At least if there is no abusable bug in my parser. But problemtools still limits, memory, ram, and execution time which is pretty important

>> timing
>
> I must admit, I'm not really up-to-date with the ctd language. Can the user write a program that's slow/hangs? I was under the assumption that it was basically just describing the shape of the data, so performance is what it is.

It could be veeery slow. Just nesting three loops (n <= 10^5) will already take minutes per testcase?
(The binary is also really slow if you use RegEx in a loop or do many `UNIQUE` calls but those issues are fixed in the python version.)

>> GIL
>
> I would have assumed the programs you generate are pretty I/O-bound, so that the GIL isn't really an issue? But relatedly, as a library, thread safety guarantees would be important.

There can be stuf that is very not IO bound. E.g. validating graphs. In those case you would loose all paralellism if you used this as a library instead of a CLI

> I'm guessing the assumption that pyctd would be installed in $PATH may be what's making this PR fail CI.

Since pyctd is installed as dependency it should be in $PATH (this also work if you use virtual environment because the virtual environment is prepended to $PATH)... I am not sure whats going wrong. The `which('pyctd')` is even superfluous just passing `pyctd` as path is supposed to work.
(btw. I am very surprised that problemtools hardcodes the locations of all compilers and does not detect them via `which`)

>> As far as I know that is not too uncommon.
>
> Can you give some examples of other projects doing this? 

BAPCtools does this 🙃 But many tools that use stuff like `mypy`, `pylint`, `Black`, `Ruff` etc. do this.

mzuenni: changing the call of checktestdata to `python -m checktestdata` seems to work better

gkreitz: OK, let's not go down the library route, but consider this a pure command line tool.

> Since pyctd is installed as dependency it should be in $PATH (this also work if you use virtual environment because the virtual environment is prepended to $PATH)... I am not sure whats going wrong. 

You can run tools directly from a venv (`venv/bin/verifyproblem`) without doing a `source activate`. This does not mess with your `$PATH`.

> (btw. I am very surprised that problemtools hardcodes the locations of all compilers and does not detect them via which)

Yeah, for compilers, we probably should look them up via `$PATH` if the path is not absolute, and default the config to be just the binary name. #103 is an old issue about this.

> BAPCtools does this 🙃 But many tools that use stuff like mypy, pylint, Black, Ruff etc. do this.

There's a huge difference to me in pulling dev dependencies and hooking them in in whatever way works vs. how we run dependencies we ship with. :)

problemtools/run/checktestdata.py
```diff
@@ -3,29 +3,24 @@
...
         """
-        if Checktestdata._CTD_PATH is None:
-            raise ProgramError('Could not locate the Checktestdata program to run %s' % path)
-        super().__init__(Checktestdata._CTD_PATH, args=[path])
+        super().__init__(sys.executable, args=['-m', 'checktestdata', path])
 
     def __str__(self) -> str:
         """String representation"""
-        return '%s' % (self.args[0])
+        return '%s' % (self.args[-1])
```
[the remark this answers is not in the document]
mzuenni: Just for completeness: pyctd is a drop in replacement and matches the exit codes of the binary

gkreitz (approved):

PR #451, opened 2026-08-06
Kill forked submission processes after execution

### LINKED ISSUE #450: If a TLE submission forks, its children don't get killed
This shows up in hacked multipass problems, where the `include` runner forks the submission.

Sakshamm-Goyal:
## Summary

- run each submission in its own session/process group
- kill any descendants left behind after the direct submission process exits
- add a regression that forks a sleeping child, simulates a CPU-limit signal, and verifies the child does not survive

Closes #450

## Validation

- fail-first regression: the forked child remained alive before the process-group cleanup
- `.venv/bin/python -m pytest -q tests/test_run_program.py tests/test_run_limit.py` — 2 passed
- `.venv/bin/python -m pytest -q --ignore=tests/test_latex.py --ignore=tests/test_markdown.py --ignore=tests/test_xss.py` — 67 passed
- `.venv/bin/ruff check .`
- `.venv/bin/ruff format --check .`
- `.venv/bin/mypy --non-interactive --config-file mypy.ini -p problemtools`

The full local pytest run additionally needs `lualatex`; without it, the first PDF-render test fails and leaves the process in a removed temporary working directory, which cascades into 10 unrelated render-test failures.

### DIFF
--- a/problemtools/run/program.py
+++ b/problemtools/run/program.py
@@ -97,6 +97,9 @@ class Program(ABC):
         pid = os.fork()
         if pid == 0:  # child
             try:
+                # Keep the submission and everything it forks in a group that
+                # can be cleaned up after the submission process exits.
+                os.setsid()
                 # The Python interpreter internally sets some signal dispositions
                 # to SIG_IGN (notably SIGPIPE), and unless we reset them manually
                 # this leaks through to the program we exec. That can has some
@@ -133,6 +136,11 @@ class Program(ABC):
             log.error('Unreachable part of run_wait reached')
             os.kill(os.getpid(), signal.SIGTERM)
         (pid, status, rusage) = os.wait4(pid, 0)
+        # The process waited for above may have left descendants behind.
+        try:
+            os.killpg(pid, signal.SIGKILL)
+        except ProcessLookupError:
+            pass
         return status, rusage.ru_utime + rusage.ru_stime
 
     @staticmethod
--- a/problemtools/run/program.py
+++ b/problemtools/run/program.py
@@ -97,9 +97,11 @@ class Program(ABC):
         pid = os.fork()
         if pid == 0:  # child
             try:
-                # Keep the submission and everything it forks in a group that
-                # can be cleaned up after the submission process exits.
-                os.setsid()
+                # Keep the submission and everything it forks in a process
+                # group that can be cleaned up after the submission exits.
+                # Unlike setsid(), this does not detach the submission from
+                # its parent session or terminal.
+                os.setpgrp()
                 # The Python interpreter internally sets some signal dispositions
                 # to SIG_IGN (notably SIGPIPE), and unless we reset them manually
                 # this leaks through to the program we exec. That can has some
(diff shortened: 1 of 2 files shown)


problemtools/run/program.py
```diff
@@ -133,6 +136,11 @@ def __run_wait(argv, infile, outfile, errfile, timelim, memlim, working_director
             log.error('Unreachable part of run_wait reached')
             os.kill(os.getpid(), signal.SIGTERM)
         (pid, status, rusage) = os.wait4(pid, 0)
+        # The process waited for above may have left descendants behind.
+        try:
+            os.killpg(pid, signal.SIGKILL)
```
simonlindholm, line 161: This will fail if there are no processes left in the process group, and may kill unrelated processes if the pid index wraps around. While hitting this wraparound is rather unrealistic in practice, we could avoid it by adding `WNOWAIT` to the wait4 call and then doing a wait again after the pgkill. (If we don't do that, it definitely deserves a comment.)

problemtools/run/program.py
```diff
@@ -97,6 +97,9 @@ def __run_wait(argv, infile, outfile, errfile, timelim, memlim, working_director
         pid = os.fork()
         if pid == 0:  # child
             try:
+                # Keep the submission and everything it forks in a group that
+                # can be cleaned up after the submission process exits.
+                os.setsid()
```
simonlindholm: os.setpgrp() feels like a less heavy hammer.

tests/test_run_program.py
```diff
@@ -0,0 +1,64 @@
...
+    try:
+        if Path(f'/proc/{pid}/stat').read_text().split()[2] == 'Z':
+            return False
+    except OSError:
+        pass
+    return True
+
+
+def test_run_kills_forked_children_after_time_limit(tmp_path, monkeypatch):
+    monkeypatch.setattr(program.limit, 'try_limit', lambda *_args: None)
```
simonlindholm: Why this monkeypatching? Feels like it just makes the test more fragile

tests/test_run_program.py
```diff
@@ -0,0 +1,64 @@
...
+from problemtools.run import program
+
+
+def _process_exists(pid: int) -> bool:
+    try:
+        os.kill(pid, 0)
+    except ProcessLookupError:
+        return False
+    try:
+        if Path(f'/proc/{pid}/stat').read_text().split()[2] == 'Z':
```
simonlindholm: This check feels wrong to me, zombies exist too in some sense and are problematic to leave around

tests/test_run_program.py
```diff
@@ -0,0 +1,64 @@
...
+            child_pid = os.fork()
+            if child_pid == 0:
+                with open({str(child_pid_file)!r}, 'w') as child_pid_output:
+                    child_pid_output.write(str(os.getpid()))
+                while True:
+                    time.sleep(60)
+
+            while not os.path.exists({str(child_pid_file)!r}):
+                time.sleep(0.01)
+            os.kill(os.getpid(), signal.SIGXCPU)
```
simonlindholm, line 52:
```suggestion
            # Speed up the test by raising a timeout signal early
            os.kill(os.getpid(), signal.SIGXCPU)
```

tests/test_run_program.py
```diff
@@ -0,0 +1,64 @@
...
+
+            child_pid = os.fork()
+            if child_pid == 0:
+                with open({str(child_pid_file)!r}, 'w') as child_pid_output:
+                    child_pid_output.write(str(os.getpid()))
+                while True:
+                    time.sleep(60)
+
+            while not os.path.exists({str(child_pid_file)!r}):
+                time.sleep(0.01)
```
simonlindholm, line 51: There's a race here where the pid file may have been created but not yet received its contents

PR #463, opened 2026-08-20
Default to looking for tools on PATH, rather than requiring them to be in `/usr/bin`

### LINKED ISSUE #103: languages.yaml should accept compilers in path and not hardwire absolute paths

gkreitz:
This PR allows one to configure a language where the path to the compiler or interpreter is not absolute (e.g., `gcc` rather then `/usr/bin/gcc`). Such paths will be resolved using `$PATH`. This PR also changes the default config to make use of this, so now it will compile C programs using the first `gcc` in `$PATH`, rather than `/usr/bin/gcc`.

Additionally, fixes an old omission. If the runtime for a language (e.g., `java`) is missing, earlier we'd attempt to execute it and give a misleading run-time error. Now, we'll flag that as a compile error, the same way we do if the compiler is missing.

Fixes #103

### DIFF
--- a/.bandit
+++ b/.bandit
@@ -1,0 +1,10 @@
+# We don't really use bandit ourselves, but codefactor runs it for us.
+# This config makes it less silly
+[bandit]
+# Plain directory names here (e.g. "tests") silently fail to exclude
+# anything, despite being the exact example bandit's docs uses...
+# https://bandit.readthedocs.io/en/latest/config.html#bandit-settings
+exclude = ./tests/*
+# Bandit warns about executing programs, using asserts, and also thinks
+# multi-pass is a hard-coded password...
+skips = B101,B105,B311,B404,B601,B603,B606,B607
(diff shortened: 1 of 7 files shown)


gkreitz:
Codefactor checks are silly (hard coded paths to /tmp in a unit tests are not a security vulnerability. I'll see if I can coax codefactor into being less silly (but iirc, it was a bit annoying to get it to pick up bandit rules in a good way).

pehrsoderman (approved):

PR #479, opened 2026-08-28
Warn for typo'd folders #399

### LINKED ISSUE #399: Warn for typo'd folders
For example, I get no warning for having a folder named `attachment`

gkreitz:
Adds a check looking at directories in the root of the problem package, warning for unknown directories (and suggesting the closest known match, if anything is close). This check replaces 3 bespoke checks for the same thing in various places.

Drops the check for unknown directories in submissions/. I covered this warning when extracting submission checks (we now look in all directories for submissions, and will warn about unknown submissions), so it's now basically a duplicate.

Renames the variable/function parameter `format` to `format_version` to avoid shadowing the built-in.

Fixes #399


pehrsoderman (approved):

simonlindholm:
This feels a bit overaggressive to me, adding warnings that you can't silence without (IMO) a clear upside (I can't think of a bug that this would have caught).

For example, for Swedish IOI qualifiers we regularly use a directory `/manual-tests/` where we put handcrafted testcases that `/data/generator.sh` then copies into `/data/`.

I've also in the recent past used directories like `old_submissions/` for trashing submissions that don't fit in `submissions/` (e.g. because of unstable verdicts, or because we changed the problem but didn't update solutions), and kept various temp folders around for dev purposes.

Tagl:
Yeah I thought the idea was to catch levenshtein distance 1 or 2 typo errors, not warn on arbitrary directories.
This will be rather annoying in current form for any sort of development.

gkreitz:
> without (IMO) a clear upside (I can't think of a bug that this would have caught).

I think the upside is exactly the example given in the issue, catching stuff like calling a directory `attachment` instead of `attachments` (which is very easy to miss).

But, good points. I'll look at tweaking to allow for more clutter.

PR #484, opened 2026-09-01
Add more warnings about score ranges (and reject_score)

### LINKED ISSUE #439: Compute score bounds based on grading conf and warn about discrepancies
- Check that no test case gets score.txt and a verdict other than accepted
- Warn if some test data node has infinite max, but we can prove from children's ranges with aggregation that a finite bound can be found
- Warn if any test data node has negative range lower bound. In particular, this will give a warning for every node without an explicitly set range, which I think is good?
- In the new "Kattis" mode, error on objective: min

gkreitz:
In these new checks, we compute max scores for test data groups based on the configuration of their children, and add several new warnings:

- If the range can be narrowed (e.g,. 10 points per test case, 7 test cases, but your range says 0 100)
- If no range was specified, but we can compute one
- If no range is specified on the root node (where you basically always want to specify something)
- If the root node has a negative minimum (which is almost always a mistake)

I decided not to add warnings if the computed range is *larger* than the declared range (e.g., 11 test cases worth 10 each, but the range says 0 100). This should be caught by out-of-bounds checks when running submissions. Happy to take opinions on if we want to check it statically here too.

Note that these checks all disappear when a custom grader is used (for obvious reasons). And if custom scoring is used, we can't make any assumptions regarding what each test case is worth, so the checks are less likely to catch stuff.

I also added an explicit warning for using `reject_score`. It's been removed in the new format. I cannot think of a good use case, and it's always the default 0 in all of the problems ever installed on Kattis.

Fixes #439


problemtools/checks/testdata.py
```diff
@@ -64,6 +75,127 @@ def check_testdata(
...
+    elif min_score < agg_min or max_score > agg_max:
+        if is_default_range:
+            diag.warning(
+                f'No score range declared for {group}, but a range of [{agg_min:g}, {agg_max:g}] can be '
+                f"computed from its grading configuration and test data; consider adding 'range: {agg_min:g} {agg_max:g}'"
+            )
+        else:
+            diag.warning(
+                f"Declared score range '{score_range}' for {group} is looser than the computed range "
+                f'[{agg_min:g}, {agg_max:g}]; consider tightening it'
```
pehrsoderman: Uses :g (usually 6 significant digits) which is inconsistent with the default grader which uses %f (6 decimals). This makes for a somewhat misleading error message: 

Assume that the default grader gives the score of `33.333333`:

```
  WARNING Declared score range '0 100' for testcase group data.secret is looser than
          the computed range [0, 33.3333]; consider tightening it
```

And if you then `range: 0 33.3333` as instructed: 

```
  ERROR submission sol.py (Python 3 (w/PyPy3)) got score 33.333333 on group data/secret,
        which is outside of expected score range [0.0, 33.3333]
```

problemtools/checks/testdata.py
```diff
@@ -64,6 +75,127 @@ def check_testdata(
...
+        )
+
+
+#: Score aggregators for `grading: default`, matching support/default_grader's `score_aggregators`.
+#: All are monotonic non-decreasing in each argument, which is what makes _check_score_range below
+#: correct: the range of an aggregate is the aggregator applied to the children's lower bounds, and
+#: separately to their upper bounds.
+_SCORE_AGGREGATORS: dict[str, Callable[[list[float]], float]] = {
+    'sum': sum,
+    'avg': lambda scores: sum(scores) / len(scores),
```
pehrsoderman: I am trying to wrap myself around if this really works. Assume we are using `grader_flags: avg, on_reject: break` and subtasks worth 100/0/0, and a submission that fails on the second subtask. Won't that give us a score of 50, instead of 33.33333, as the third subtask is not scored?

pehrsoderman (approved):
Two comments to review before merge.

problemtools/checks/testdata.py
gkreitz: Good catch, `avg` is a bit more annoying than I realized.

problemtools/checks/testdata.py
gkreitz: This does not feel like an actual issue. It's going to be exceedingly rare that you end up with a suggestion with decimals. If a human does, and copy-pastes, they will swiftly see the problem add add more decimals to the range. I do not think it's worth it trying to adjust precision to what we use for the grader.

problemtools/checks/testdata.py
```diff
@@ -64,6 +75,141 @@ def check_testdata(
...
+
+def _warn_score_range(
+    group: TestDataGroup, declared: tuple[float, float], aggregate: tuple[float, float], diag: Diagnostics
+) -> tuple[float, float]:
+    """Compare `group`'s declared score range to what its `aggregate` says can actually be achieved,
+    warn about any mismatch, and return the effective (declared-trusting) range."""
+    min_score, max_score = declared
+    agg_min, agg_max = aggregate
+    score_range = group.config['range']
+    is_default_range = score_range == DEFAULT_CONFIG['range']
```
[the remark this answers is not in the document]
gkreitz: This is used as a heuristic to detect the value being unconfigured, so it *should* be brittle. Rewriting testdatagroup configuration is on my list, though, but not part of this PR.

PR #485, opened 2026-09-02
Try to be smarter about when to warn for ans files not being accepted

### LINKED ISSUE #344: Improve warning for when the `ans` file of a test case is not accepted by the validator
We currently run the validator on every `ans` file provided, and emit a warning (or error, if in `sample/`) if the test case is not accepted. The logic is a bit flawed, but we'd need to restructure things a bit to get a clean way to express checks where the logic depends on the result across all test cases (we currently build up a class hierarchy very closely resembling the directory tree).

My impression is that it's typically the case that the answer file is either:

1. The same format as what submissions should output.
2. A completely different format with some auxiliary data needed by the output validator - here, *every* ans file is expected to get WA, so the current warning gets super spammy.

There's no way in the problem format package to indicate which of the two worlds we're in, but better logic for this check would be:
- If the answer file gets a WA for *all* test cases, just add a notice.
- If the answer file gets a WA for *any* test case in sample, that's an error (this part of the current check is good).
- If the answer file gets a JE for *any* test case, that's an error.
- If the answer file gets a WA for *some* test cases, but AC for others, that's a very strong warning (I'd love to make this an error, but this is technically fine, and could be valid if the validator only sometimes needs auxiliary data and the problem creator took care to avoid it for e.g. samples)

gkreitz:
Changes our warning when `.ans` files are not accepted by the output validator, in line with what I outlined in #344.

When looking at secret test cases, our logic now looks at the validation result across all of them. If the output validator rejects the `.ans` file on all of them, we just do an `.info` (not shown by default), as this means we're very likely in a scenario where the `.ans`  file should not look like the output. If some get AC and some don't, we give a (hopefully clear) warning. We only give *one* warning now, not one per test case.

Fixes #344


problemtools/checks/testdata.py
```diff
@@ -254,20 +248,57 @@ def _check_testcase(
...
+
+    if rejected := [testcase for testcase, val_res in results if val_res.verdict != 'AC']:
+        if any(val_res.verdict == 'AC' for _, val_res in results):
+            diag.warning(
+                f'judge answer file was not accepted by the output validator on {len(rejected)}/{len(secret)} secret '
+                f'testcases (e.g. testcase {rejected[0].path}); this is fine if the answer files intentionally use a '
+                'different format than what is expected from submissions, but is suspicious when only some of them do'
+            )
+        else:
+            diag.info(
```
pehrsoderman: So, if the filtered list of testcases are all rejected, we downgrade from .warning to .info. I think this interacts somewhat badly with `--data_filter`, as filtering out a set of datafiles where all are rejected will result in a .info message, that is not displayed by default(?) 

Similar issue if there is only a single file, in which case there will at most be a .info message.

pehrsoderman (approved):

problemtools/checks/testdata.py
```diff
@@ -254,20 +248,57 @@ def _check_testcase(
...
+                f'judge answer file was not accepted by the output validator on any of the {len(secret)} secret '
+                'testcases; this is fine if the answer files intentionally use a different format than what is '
+                'expected from submissions'
+            )
```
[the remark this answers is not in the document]
gkreitz: It's fine to be a bit spammy if we trigger JE.

problemtools/checks/testdata.py
gkreitz: Taken in the context of problem development, I don't think the corner cases you bring up matter.  Feel free to suggest other strategies for this warning if you disagree. I think this PR, as written is a clear improvement on the status quo.

Note that the old warning was extremely spammy, and likely did more harm than good (the warning fatigue from warning once per file about a legitimate, and relatively common, setup vs. catching something which is honestly extremely hard to do by accident).

PR #486, opened 2026-09-02
Surface image errors and warnings from HTML rendering in verifyproblem

### LINKED ISSUE #470: pdf -> png conversion can fail silently
If `gs` fails to convert a pdf to png when rendering to html, verifyproblem will not give an error or even a warning. The error handling in that path seems to be a bit of a mess.

gkreitz:
Starts threading in Diagnostics into PlasTeX rendering to be able to better flag errors and warnings during tex -> html rendering. Also adds a check when converting pdf images to png, catching `gs` exiting with 0 without producing any png file.

After this PR, if rendering tex->html causes warnings/errors, those will be spotted by verifyproblem (and not just the generic failure message telling you to run problem2html to diagnose, even if that can still happen). This means that having large images is now a warning directly in verifyproblem, and failing to convert images is an error. I think this is a clear step up, but happy to hear feedback on if this ends up too spammy for some problems.

As a bonus, errors/warnings we ourselves emit in problem2html for tex->html are now colorful, like in verifyproblem.

This PR also deprecates `--quiet` in problem2html. That switch now only controls 2 print statements (use `--log-level` instead to control how much output you want). For obvious reasons, `--quiet` will not print a deprecation warning. :)

Fixes #470


Matistjati:
I tested it on ~10 problems from PO where we have illustrations. No auto-generated illustrations triggered the warning, but real photos frequently triggered the uppercase warning. I think this is fine, since I have yet to see a problem where all that resolution is actually used. Perhaps this could've saved us from the 20k x 20k picture here 🙂 https://open.kattis.com/problems/investigatingfrogbehaviouronlilypadpatterns

gkreitz:
Great, thanks for testing!

The uppercase warning has in some sense always been there (if one manually ran problem2html). I take it from your comment, that it's probably the right call to surface it in verifyproblem (I'm 99% sure the reason we hid it had more to do with earlier implementation limitations, rather than a choice)

PR #493, opened 2026-09-23
Set max-width in problem2html css

Matistjati:
In my experience, people don't really use problem2html because it looks very different from Kattis because we have no max-width set. This becomes especially bad if you have a wide monitor.

bild

So let's set some max-width. As far as I'm aware, the only two CCS that use poblemtools and render to HTML are Kattis and omogenjudge, both of which set a max-width similar to LaTeX. So let's render something closer to what you actually get.

bild


Tagl:
I would like to note however that the width on Kattis is smaller than what you get in the PDF. It is in fact very very small.
You can see an example here: https://open.kattis.com/problems/dialhomedevice
The table fits in the PDF but doesn't on the web.

